PR #1, opened 2025-07-02
Adjustment for js/tsx frontend analysis

yhoztak:
Modified frontend_security.ron so that it has high accuracy of findings. 
- Dropping some function from sanitizers as they don't really sanitize it and still raises risks. 

- Added logic for tsx file to look at javascript rules. 

- Added code type detection (frontend /backend) and language filter

- fixed taint system for javascript. some declaration were missing and also not supporting js style declaration like `const/let/var variable = value`

- Updated rule for Prototype Pollution

image

### DIFF
--- a/src/scanner/utils.rs
+++ b/src/scanner/utils.rs
@@ -70,6 +70,9 @@ pub fn rule_applies_to_file_path(file_types: Option<&FileTypes>, file_path: &Pat
 
 /// Detect programming language from file path
 pub fn detect_language_from_path(file_path: &Path) -> Option<&'static str> {
+    let file_name = file_path.file_name()?.to_str()?;
+    
+    // Handle standard extensions
     match file_path.extension()?.to_str()? {
         // Python extensions
         "py" | "pyw" | "pyi" | "pyx" => Some("python"),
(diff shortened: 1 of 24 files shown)


src/scanner/utils.rs
```diff
@@ -333,19 +333,22 @@ impl AstUtils {
...
         match language {
-            "javascript" | "typescript" => Self::check_javascript_sanitization(code),
+            "javascript" | "typescript" => Self::check_html_sanitization(code),
             "python" => Self::check_python_sanitization(code),
             "java" => Self::check_java_sanitization(code),
             _ => Self::check_generic_sanitization(code),
         }
     }
 
-    fn check_javascript_sanitization(code: &str) -> bool {
```
asadeddin, line 343: @yhoztak this is fine for now, but do we want to move this logic to be defined on the rules rather than here? This is technical debt IMO. What do you think?

asadeddin (approved):

src/scanner/utils.rs
yhoztak: yes, taint rules already have sanitizers so I think it make sense to have it under rules. I'll look into this later to see if we can remove these

test_files/false_positives/cleartext_transmission_false_positives.js
```diff
@@ -0,0 +1,223 @@
...
+    var password = document.getElementById('password').value;
+    var xhr = new XMLHttpRequest();
+    xhr.open('POST', 'http://example.com/login'); // HTTP not HTTPS
+    xhr.send(JSON.stringify({ password: password }));
+}
+
+// 2. Real form data with sensitive token transmitted over HTTP
+function submitTokenForm() {
+    var userToken = document.querySelector('input[name=token]').value;
+    fetch('http://api.example.com/data', { // HTTP not HTTPS
```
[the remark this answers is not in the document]
yhoztak: These should be ignored by fusion ignore_paths actually

PR #2, opened 2025-07-04
Supporting .gitignore to ignore files

Ibrahimrahhal:
I was playing around today with this CLI but it was very slow and noisy locally because it was including all minified and ignored files.


yhoztak:
I also realized that some exclusion patterns are not working as intended for js files and they are fixed as part of https://github.com/Corgea/Sighthound/pull/1

yhoztak (approved):
LGTM

PR #3, opened 2025-07-08
Add CWE as response, export linux compatible binary, etc for fusion integration

yhoztak:
While integration on fusion, made the following changes

- embed rules as part of binary so that we don't have to export along with rules everytime.
- added cwe_id field as part of response so that consumer of this doesn't have to map it, or guess which tag to use.
- Added Dockerfile that we can export linux compatible binary
- ruff format


asadeddin (approved):
LGTM

PR #4, opened 2025-07-10
rules update, combine simple search & taint analysis as default behavior

yhoztak:
- update rules to reduce false negatives
- combine simple search and taint analysis as default behavior, keeping the option to run each 
- adding `--version` to check what the current version of the binary

updated comparison result: 
https://docs.google.com/spreadsheets/d/1Hu56oyHm6E1jt7yN8JL_2NpC98OJhXpOGrYxw_jD7bs/edit?gid=1638247948#gid=1638247948


asadeddin (approved):
LGTM

PR #9, opened 2026-01-21
Html support

yhoztak:
Adding html, sql, config, properties support for detecting simple vulnerabilities.


rules/config/config_security.ron
```diff
@@ -0,0 +1,233 @@
+(
+    rules: [
+        (
+            id: Some("config-secret-001"),
+            name: Some("Hardcoded Password in Config"),
```
abronte: Doesn't this already get handled by our secret scanning?

abronte:
Overall, LGTM but I don't think sighthound should be detecting anything hardcoded credentials related.

rules/xml/xml_security.ron
```diff
@@ -0,0 +1,184 @@
...
+        (
+            id: Some("xml-xxe-001"),
+            name: Some("XML External Entity (XXE) - DOCTYPE Declaration"),
+            category: Some("xxe"),
+            mode: "search",
+            patterns: Some([
+                "<!DOCTYPE*<!ENTITY",
+                "<!DOCTYPE*SYSTEM",
+                "<!ENTITY*SYSTEM",
+            ]),
```
0xjgv, line 12: These patterns use `*` as an inline wildcard, but the matcher treats it as an anchored glob over the entire file text. `<!DOCTYPE*SYSTEM` therefore does not match normal XML files, and the same issue affects several new SQL and HTML rules. Use plain substrings or `regex:` patterns instead.

rules/xml/xml_security.ron
```diff
@@ -0,0 +1,184 @@
...
+            patterns: Some([
+                "http://",
+            ]),
+            unless: Some([
+                "http://localhost",
+                "http://127.0.0.1",
+                "http://www.w3.org",
+                "http://schemas.",
+                "xmlns:*=\"http://",
+            ]),
```
0xjgv, line 54: `UnifiedRule` has no `unless` field, so serde drops these exclusions and nothing evaluates them. `xml-insecure-001` will flag benign namespace URLs despite this block; the same applies to the new HTML and config exclusions. Add matcher support before relying on `unless`, or remove these clauses.

Cargo.toml
```diff
@@ -11,13 +11,17 @@ name = "sighthound"
...
 python = ["tree-sitter-python"]
 java = ["tree-sitter-java"]
 javascript = ["tree-sitter-javascript"]
 tsx = ["tree-sitter-typescript"]
 html = ["tree-sitter-html"]
 django = ["tree-sitter-html"]  # Reuse HTML parser for Django templates
+sql = []  # Text-based pattern matching, no parser needed
+xml = []  # Text-based pattern matching, no parser needed
+properties = []  # Text-based pattern matching, no parser needed
+config = []  # Text-based pattern matching, no parser needed
```
0xjgv: should `sql`, `xml`, `properties`, and `config` enable `tree-sitter-javascript` or avoid that parser? With `--no-default-features --features sql`, Cargo omits the optional crate while these implementations reference it unconditionally, so that feature combination does not compile.

tests/strictness/main.rs
```diff
@@ -10,6 +10,7 @@ mod cross_file_taint;
 mod django_security;
 mod false_positive_regressions;
 mod language_coverage;
+mod markup_config_languages;
```
juangaitanv, line 13: `strictness_tests` cannot compile because `markup_config_languages` has no module. Could we add the intended test module or remove the declaration?

src/scanner/utils.rs
```diff
@@ -275,6 +275,16 @@ pub fn detect_language_from_path(file_path: &Path) -> Option<&'static str> {
...
         | "jade" => Some("html"),
 
+        "sql" | "ddl" | "dml" => Some("sql"),
+
+        "xml" | "xsd" | "xsl" | "xslt" | "wsdl" | "svg" | "pom" => Some("xml"),
+
+        "properties" | "props" => Some("properties"),
+
+        "conf" | "cfg" | "ini" | "env" | "config" | "json" | "yaml" | "yml" | "toml" => {
+            Some("config")
```
juangaitanv, line 285: the routed extensions do not satisfy the loaded rules' `file_types`; could we align routing, rule packs, and `file_types` so they receive the advertised coverage?

src/scanner/scanning_logic.rs
```diff
@@ -188,14 +226,218 @@ impl ScanningLogic {
...
+        findings: &mut Vec<crate::models::Finding>,
+        processed: &mut std::collections::HashSet<FindingDedupKey>,
+    ) {
+        for line in lines {
+            if line.searchable.trim().is_empty()
+                || !crate::rules::rule_matches_pattern_scoped(
+                    rule,
+                    &line.searchable,
+                    &line.searchable,
+                )
```
juangaitanv, line 281: a matching `unless` value elsewhere on the line suppresses later positive matches; should we evaluate exclusions per match or local context?

rules/html/html_security.ron
```diff
@@ -0,0 +1,227 @@
...
+        (
+            id: Some("html-xss-001"),
+            name: Some("JavaScript URL in href Attribute"),
+            category: Some("xss"),
+            mode: "search",
+            patterns: Some([
+                "href=\"javascript:",
+                "href='javascript:",
+                "href=javascript:",
+            ]),
```
juangaitanv, line 12: these patterns miss case and whitespace variants and can match names like `data-href`; could we use boundary-aware, case-insensitive attribute matching?

rules/config/config_security.ron
```diff
@@ -0,0 +1,103 @@
...
+            name: Some("Debug Mode Enabled in Config"),
+            category: Some("configuration"),
+            mode: "search",
+            patterns: Some([
+                "\"debug\": true",
+                "'debug': true",
+                "debug: true",
+                "\"DEBUG\": true",
+                "DEBUG: true",
+            ]),
```
juangaitanv, line 52: these patterns miss compact JSON and TOML `=` syntax; could we allow whitespace variation and both `:` and `=` separators?

src/scanner/scanning_logic.rs
```diff
@@ -188,14 +226,218 @@ impl ScanningLogic {
...
+    }
+
+    fn is_text_comment_line(line: &str, language: &str) -> bool {
+        let trimmed = line.trim_start();
+        match language {
+            "sql" => trimmed.starts_with("--") || trimmed.starts_with('#'),
+            "properties" => trimmed.starts_with('#') || trimmed.starts_with('!'),
+            "config" => {
+                trimmed.starts_with('#') || trimmed.starts_with(';') || trimmed.starts_with("//")
+            }
```
juangaitanv, line 428: inline YAML and TOML comments remain searchable and can produce findings; could we strip inline comments while preserving markers in quoted values?

juangaitanv:
Superseded in part by #75.

That PR ports the same 12 HTML rule IDs — identical rule-ID set to this branch's `rules/html/html_security.ron`, verified by sorted-ID diff. It also carries a 21-line engine-notes comment block this branch does not have.

Everything else here is unique to this PR and is **not** carried over: `rules/{config,properties,sql,xml}/*.ron` (~323 lines of new rules), the parser-free text-scanning engine (`scan_text_file_with_rules` and helpers, `SearchSemantics::Text`), and the `unless` exclusion mechanism. Those were all explicitly out of scope for the HTML chunk, so this PR should stay open for them.

Three things worth knowing independently:

1. **This branch is 11 commits stale.** It was cut from `80cf0dc`, not current `main` — `git merge-base --is-ancestor 529cbd7 f73718d3` returns 1. A plain `git diff main...` overstates its surface by 9 files that are pure drift (`src/scanner/ast_provenance.rs`, `dataflow.rs`, `flow_tracker.rs`, and others). Worth a rebase before review.

2. **`strictness_tests` does not compile at this head.** `tests/strictness/main.rs` declares `mod markup_config_languages;` and no file of that name exists in the tree or in reachable history. `Cargo.toml` maps `[[test]] name = "strictness_tests"` to that file, so the target fails module resolution — meaning the strictness tests for the sql/xml/config/properties work cannot currently run at this SHA.

3. **The `unless` work fixes a live bug.** `rules/javascript/frontend_security.ron` already ships three `unless:` blocks on `main` today, and nothing reads them: `UnifiedRule` has no such field and there is no `deny_unknown_fields`, so serde drops them silently. Those exclusions are non-functional right now. The `unless` support in this PR is what makes them take effect.

juangaitanv:
Superseded in part by https://github.com/Corgea/Sighthound/pull/75

PR #21, opened 2026-06-15
Prepare open-source release (license, CI, community docs)

asadeddin:
## Summary

- Add Apache-2.0/MIT dual license, CI and release workflows, contributing and security policies
- Refresh README, Dockerfile, Makefile, and build tooling
- Add `examples/basic_scan.rs`

**Merge order:** PR 1 of 3 remaining — merge first after #16–#19 (already on `feature/more_language_support`). Docs/CI only — no engine changes.

**Note:** Rebased onto current `feature/more_language_support` (includes `tree-sitter 0.26` C# fix). PR #20 (benchmark harness) was closed — no `bench/` tooling in this stack.

## Test plan

- [ ] CI workflow passes
- [x] `cargo build`
- [x] `cargo test`
---
**Integration branch:** Merges into [`feature/more_language_support`](https://github.com/Corgea/Sighthound/tree/feature/more_language_support) — **not** `main`. When the full stack is landed on the feature branch, we will open one PR from `feature/more_language_support` → `main`.

**Merge order:** #16–#19 (merged) → **#21** → #22 → #23

PR #20, opened 2026-06-15
Add benchmark harness for Sighthound vs Semgrep/OpenGrep

asadeddin:
## Summary

- Add `bench/score.py`, `bench/run_bench.py`, and `bench/compare.py` for CWE-scored benchmarking
- Add comparison documentation and bench README

**Merge order:** PR 5 of 8 — merge after PR 2 (parallel with PR 3 and PR 4). Recommended last in wave 3.

## Test plan

- [ ] `python3 bench/run_bench.py --help`
---
**Integration branch:** Merges into [`feature/more_language_support`](https://github.com/Corgea/Sighthound/tree/feature/more_language_support) — **not** `main`. When the full stack is landed on the feature branch, we will open one PR from `feature/more_language_support` → `main`.

**Merge order:** #16 → #17 → #18/#19/#20 (parallel OK) → #21 → #22 → #23

PR #19, opened 2026-06-15
Add HttpClient SSRF search rule for C#

asadeddin:
## Summary

- Replace the URL-build taint SSRF rule with search patterns for `GetStringAsync`, `GetByteArrayAsync`, and `GetStreamAsync`
- C# call-name resolution landed in PR #17 — this PR is rules-only

**Merge order:** PR 4 of 8 — merge after PR 2 (parallel with PR 3 and PR 5).

## Test plan

- [x] `cargo test`
- [ ] C# HttpClient SSRF rule matches method calls
---
**Integration branch:** Merges into [`feature/more_language_support`](https://github.com/Corgea/Sighthound/tree/feature/more_language_support) — **not** `main`. When the full stack is landed on the feature branch, we will open one PR from `feature/more_language_support` → `main`.

**Merge order:** #16 → #17 → #18/#19/#20 (parallel OK) → #21 → #22 → #23

PR #16, opened 2026-06-15
Add tree-sitter grammar features and LanguageSupport for expanded languages

asadeddin:
## Summary

- Add optional tree-sitter grammar Cargo features for Go, Ruby, C#, TypeScript, HTML, and PHP
- Implement `LanguageSupport` trait for expanded languages in `src/language.rs`
- Fix C# `invocation_expression` call-name resolution via the `function` field (required for method-name search rules)

**Merge order:** PR 1 of 8 — merge first. Blocks all subsequent split PRs.

## Test plan

- [x] `cargo build`
- [ ] CI passes on merge

Note: full `cargo test` suite is ported to `sighthound` crate naming in PR #17.
---
**Integration branch:** Merges into [`feature/more_language_support`](https://github.com/Corgea/Sighthound/tree/feature/more_language_support) — **not** `main`. When the full stack is landed on the feature branch, we will open one PR from `feature/more_language_support` → `main`.

**Merge order:** #16 → #17 → #18/#19/#20 (parallel OK) → #21 → #22 → #23

PR #22, opened 2026-06-15
Modernize scanner engine (taint propagation, prefilter, parallel scan)

asadeddin:
## Summary

Major scanner engine refactor. **~4k lines changed, but a large share is `rustfmt`/import reordering in tests** — the functional changes are concentrated in `src/scanner/` and the new library/CLI surface.

**Merge order:** PR 1 of 2 — merge first. Rebased directly onto `feature/more_language_support`. PRs #20 (benchmark harness) and #21 (OSS release prep) were closed.

---

### Taint analysis (`src/scanner/core.rs` — bulk of the diff)
- **Assignment-based taint tracking** — records sources from assignment RHS values, not just direct source-pattern matches on call nodes (`scan_assignments`, `scan_node_for_assignments`)
- **Propagation through dependencies** — tracks variable-to-variable flow within a function (`record_tainted_variable`, dependency extraction)
- **Sanitizer-aware sink checks** — skips findings when sanitizers appear in the sink expression
- **Flow deduplication** — `is_flow_processed` / `mark_flow_processed` to avoid duplicate source→sink reports
- **DOM assignment detection** — flags JS patterns like `.innerHTML =` as taint-relevant assignments
- **Cross-file taint** — improved import resolution and multi-file chain analysis; **skips cross-file for frontend scans** (performance)
- Richer `TaintTrace` metadata (`operation`, `function`, `rule_name`)

### Scan orchestration (`src/scanner/modes.rs`, `src/scanner/utils.rs`)
- Refactored scan entry points and parallel file discovery
- Unified/simple/taint mode wiring cleanup

### Prefilter (`src/scanner/prefilter.rs`)
- Improved rule prefiltering before full AST walks

### Pattern matching (`src/common.rs`)
- `regex:` prefix support, `*.property` wildcard matching
- Better assignment variable extraction (`extract_variable_from_assignment`)
- Literal detection improvements (quoted strings, `__all__`, etc.)

### Code-type detection (`src/code_type_detector.rs`)
- New `FrameworkRegistry` with frontend/backend framework signatures (React, Next.js, Django, etc.)
- `detect_code_type()` for frontend vs backend classification

### Rules loading (`src/rules.rs`)
- Embedded JavaScript **backend** rules (`EMBEDDED_JAVASCRIPT_BACKEND`)
- Clearer embedded-vs-file rule loading and error messages

### CLI & library surface (`src/main.rs`, `src/lib.rs`, `src/models.rs`, `src/ui.rs`)
- **New public library API** — exports `VulnerabilityScanner`, `run_auto_detection_scan`, `run_taint_analysis`, etc.
- New CLI flags: `--format` (text/json/csv), `--summary-only`, `--taint-analysis`, `--code-type`, `--language`, `--rules-dir`, `--use-embedded-rules`, `--no-parallel`, `--threads`
- New `src/ui.rs` — TTY-aware ANSI styling and spinner helpers (no emojis)
- Better invalid-argument error messages and `--version` build metadata

### Language support (`src/language.rs`)
- Mostly formatting; minor `get_function_name` cleanup for Java/JS/TS call nodes

---

## Test changes (mostly formatting)

- **~1.1k test diff lines**, but most are `rustfmt` (import order, line wrapping, signature formatting) — not new assertions
- Fixture path tweaks in `tests/test_files/python/` (comment/header lines)
- E2E injection tests updated for new scan APIs
- No new test crate — existing unit/integration/e2e suites adjusted to compile against refactored APIs

---

## Files to focus review on

| Priority | Files |
|----------|-------|
| High | `src/scanner/core.rs`, `src/scanner/modes.rs`, `src/common.rs` |
| Medium | `src/scanner/prefilter.rs`, `src/scanner/utils.rs`, `src/rules.rs`, `src/code_type_detector.rs` |
| Low (formatting) | `src/language.rs`, `tests/unit/*.rs`, `tests/integration/*.rs` |

## Test plan

- [x] `cargo test` (82 tests)

---
**Integration branch:** Merges into [`feature/more_language_support`](https://github.com/Corgea/Sighthound/tree/feature/more_language_support) — **not** `main`.

**Merge order:** #16–#19 (merged) → **#22** → #23

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,4 +1,34 @@
+//! Sighthound — a fast, AST-based static analysis scanner for security
+//! vulnerabilities in source code.
+//!
+//! Sighthound parses source with [tree-sitter](https://tree-sitter.github.io/) and
+//! applies RON-encoded rules in two modes:
+//!
+//! - **Search rules** match dangerous patterns directly in the AST.
+//! - **Taint rules** track untrusted data from sources to sinks.
+//!
+//! # Example
+//!
+//! ```no_run
+//! use anyhow::Result;
+//! use clap::Parser;
+//! use sighthound::{run_auto_detection_scan, Cli};
+//!
+//! fn scan(path: &str) -> Result<()> {
+//!     let cli = Cli::parse_from(["sighthound", path]);
+//!     let findings = run_auto_detection_scan(&cli, path, false)?;
+//!     println!("{} findings", findings.len());
+//!     Ok(())
+//! }
+//! ```
+//!
+//! # Stability
+//!
+//! The library API is **unstable** before v1.0. Minor releases may include
+//! breaking changes to public modules.
+
 pub mod cli;
+pub mod code_type_detector;
 pub mod common;
 pub mod config;
 pub mod language;
@@ -6,28 +36,23 @@ pub mod models;
 pub mod parser;
 pub mod rules;
 pub mod scanner;
-pub mod code_type_detector;
 pub mod ui;
 
-// Re-export the main types and functions that main.rs needs
-pub use scanner::{
-    VulnerabilityScanner, ScanningLogic, PreFilter, FilterStats, 
-    run_explicit_scan, run_auto_detection_scan, run_taint_analysis
-};
 pub use scanner::modes::run_taint_analysis_with_verbosity;
+pub use scanner::{
+    run_auto_detection_scan, run_explicit_scan, run_taint_analysis, FilterStats, PreFilter,
+    ScanningLogic, VulnerabilityScanner,
+};
 
-// Re-export types needed by tests and library users
 pub use common::CommonUtils;
 pub use config::ScanDefaults;
-pub use rules::{Rules, match_pattern, check_for_injection_pattern};
+pub use rules::{check_for_injection_pattern, match_pattern, Rules};
 
-// Re-export code type detection
-pub use code_type_detector::{CodeTypeDetector, CodeType};
+pub use code_type_detector::{CodeType, CodeTypeDetector};
 
-// Re-export commonly used items from models
 pub use models::{
-    Finding, TaintFlow, TaintSource, TaintSink, TaintTrace, TaintSummary,
-    UnifiedRule, FileTypes, Condition, Cli, LanguageInfo, FileInfo
+    Cli, Condition, FileInfo, FileTypes, Finding, LanguageInfo, TaintFlow, TaintSink, TaintSource,
+    TaintSummary, TaintTrace, UnifiedRule,
 };
 
 pub use parser::traverse_calls_only;
--- a/src/main.rs
+++ b/src/main.rs
@@ -104,8 +104,8 @@ fn main() -> Result<()> {
     let duration = start_time.elapsed();
 
     match cli.output_format.as_str() {
-        "json" => print_findings_json(&findings),
-        "csv" => print_findings_csv(&findings),
+        "json" => print_findings_json(&findings)?,
+        "csv" => print_findings_csv(&findings)?,
         _ => print_findings_text(&findings, cli.verbose, cli.summary_only, duration),
     }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -12,19 +12,29 @@ fn main() -> Result<()> {
     // Handle version flag
     if cli.version {
         println!("sighthound {}", env!("CARGO_PKG_VERSION"));
-        println!("Built from commit: {}", option_env!("GIT_HASH").unwrap_or("unknown"));
-        println!("Build date: {}", option_env!("BUILD_DATE").unwrap_or("unknown"));
+        println!(
+            "Built from commit: {}",
+            option_env!("GIT_HASH").unwrap_or("unknown")
+        );
+        println!(
+            "Build date: {}",
+            option_env!("BUILD_DATE").unwrap_or("unknown")
+        );
         return Ok(());
     }
 
     // Check if root_dir is provided (required for actual scanning)
     let root_dir = cli.root_dir.as_ref().ok_or_else(|| {
-        anyhow::anyhow!("Root directory is required for scanning. Use --help for usage information.")
+        anyhow::anyhow!(
+            "Root directory is required for scanning. Use --help for usage information."
+        )
     })?;
 
     // Initialize logger (respect RUST_LOG or --verbose flag)
-    env_logger::Builder::from_env(env_logger::Env::default().default_filter_or(if cli.verbose { "debug" } else { "info" }))
-        .init();
+    env_logger::Builder::from_env(
+        env_logger::Env::default().default_filter_or(if cli.verbose { "debug" } else { "info" }),
+    )
+    .init();
     // Configure threading if specified
     if let Some(threads) = cli.threads {
         rayon::ThreadPoolBuilder::new()
@@ -34,10 +44,10 @@ fn main() -> Result<()> {
     }
 
     let start_time = std::time::Instant::now();
-    
+
     // Determine if we should show progress (suppress for structured output formats)
     let show_progress = !matches!(cli.output_format.as_str(), "json" | "csv");
-    
+
     // Validate CLI parameters
     if cli.taint_analysis && cli.simple_analysis {
         return Err(anyhow::anyhow!("Cannot specify both --taint-analysis and --simple-analysis. Use one or neither (default: both modes)."));
@@ -45,7 +55,7 @@ fn main() -> Result<()> {
 
     // Resolve the actual embedded rules setting (use_file_rules overrides use_embedded_rules)
     let should_use_embedded = cli.use_embedded_rules && !cli.use_file_rules;
-    
+
     // Validate CLI parameters using CommonUtils
     CommonUtils::validate_cli_params(&cli.language, &cli.rules_path, cli.use_embedded_rules, cli.use_file_rules)
         .map_err(|e| anyhow::anyhow!(
@@ -61,36 +71,43 @@ fn main() -> Result<()> {
 
     // Handle all vulnerability scanning modes with unified flow
     let findings = if cli.taint_analysis {
+        // Only taint analysis
         run_taint_analysis(&cli, root_dir, show_progress)?
     } else if cli.simple_analysis {
+        // Only simple analysis
         match (&cli.language, &cli.rules_path, should_use_embedded) {
             (Some(_), Some(_), false) => run_explicit_scan(&cli, root_dir, show_progress)?,
             (Some(_), None, true) => run_explicit_scan(&cli, root_dir, show_progress)?,
             (None, None, _) => run_auto_detection_scan(&cli, root_dir, show_progress)?,
-            _ => unreachable!(),
+            _ => unreachable!(), // Validation above ensures this won't happen
         }
     } else {
+        // Default: Run both simple and taint analysis
+        // Run simple analysis first
         let mut simple_findings = match (&cli.language, &cli.rules_path, should_use_embedded) {
             (Some(_), Some(_), false) => run_explicit_scan(&cli, root_dir, show_progress)?,
             (Some(_), None, true) => run_explicit_scan(&cli, root_dir, show_progress)?,
             (None, None, _) => run_auto_detection_scan(&cli, root_dir, show_progress)?,
-            _ => unreachable!(),
+            _ => unreachable!(), // Validation above ensures this won't happen
         };
 
+        // Run taint analysis second (less verbose in combined mode)
         let mut taint_findings =
             run_taint_analysis_with_verbosity(&cli, root_dir, show_progress, false)?;
 
+        // Combine findings
         simple_findings.append(&mut taint_findings);
         simple_findings
     };
 
+    // Output results
     let duration = start_time.elapsed();
 
     match cli.output_format.as_str() {
         "json" => print_findings_json(&findings),
         "csv" => print_findings_csv(&findings),
-        "text" | _ => print_findings_text(&findings, cli.verbose, cli.summary_only, duration),
+        _ => print_findings_text(&findings, cli.verbose, cli.summary_only, duration),
     }
 
     Ok(())
-} 
+}
--- a/src/models.rs
+++ b/src/models.rs
@@ -52,8 +52,8 @@ pub struct TraceStep {
     pub line: usize,
     pub code: String,
     pub variable: String,
-    pub operation: String,   // "assignment", "parameter", "return", "method_call"
-    pub function: String,    // Containing function name
+    pub operation: String, // "assignment", "parameter", "return", "method_call"
+    pub function: String,  // Containing function name
 }
 
 /// Core taint flow data structure
@@ -71,7 +71,7 @@ pub struct TaintFlow {
     pub is_cross_file: bool,
     // Rule information for better reporting
     pub rule_id: Option<String>,
-    pub rule_name: Option<String>, 
+    pub rule_name: Option<String>,
     pub rule_description: Option<String>,
     pub rule_finding_type: Option<String>,
 }
@@ -98,7 +98,7 @@ pub struct TaintSink {
     pub variable: String,
     pub operation: String,
     pub code: String,
-    // Branch tracking for control flow awareness  
+    // Branch tracking for control flow awareness
     pub branch_id: Option<String>,
 }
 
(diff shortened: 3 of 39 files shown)


juangaitanv (commented):
🚀

juangaitanv (commented):
Posting the engine-refactor feedback inline. Headline thread: the cross-file taint path looks like it lost its real resolver (new `DataFlowTracer` only holds caches, never initializes import/function state), which is likely why the hardcoded `module_a.py`/`module_b.py` fallbacks remain, plus a few stdout/metadata items touch the fusion contract.

Two that land on pre-existing (unchanged) lines, so they're here rather than inline:
- `core.rs` sink/sanitizer check: should the sanitizer match confirm it wraps the tainted variable rather than appearing anywhere in the sink line?
- `core.rs` `is_dom_assignment`: does `.innerHTML =` detection need to handle a comparison in the assigned value, e.g. `el.innerHTML = a == b ? x : y`?
- `models.rs`/`main.rs`/`common.rs` nit: could we centralize `use_embedded_rules && !use_file_rules` in one helper instead of repeating it?

Verified non-issues so they don't get chased: the `detect_from_ast` tie-break rewrite is behavior-preserving; the frontend cross-file early-return doesn't over-skip single-file findings; `rules.rs`/`language.rs`/`parser.rs`/`config.rs` are pure formatting.

src/scanner/core.rs
```diff
@@ -250,9 +305,20 @@ impl ScanningLogic {
             }
         }
 
-        if language_support.name() == "javascript" || language_support.name() == "typescript" || language_support.name() == "tsx"
-            || language_support.name() == "php" {
-            Self::scan_assignments(tree.root_node(), source, filepath, rules, language_support, &mut findings, &mut processed_lines);
+        if language_support.name() == "javascript"
```
juangaitanv, line 309: should the `scan_assignments` language gate include `python`? `scan_node_for_assignments` still handles Python `assignment`/`augmented_assignment`, which looks dead for this path otherwise.

src/scanner/core.rs
```diff
@@ -286,21 +361,52 @@ impl ScanningLogic {
...
+                                finding.line,
+                                finding.function.clone(),
+                                finding.finding_type.clone(),
+                            );
+                            // A call-shaped sink (e.g. subprocess.Popen(shell=True)) can be
+                            // matched both by the call pass and here via its `=`-bearing
+                            // pattern; the call pass records a different `function`, so also
+                            // guard on (line, finding_type) to avoid a duplicate finding.
+                            let already = processed_lines.contains(&line_key)
+                                || findings.iter().any(|f| {
```
juangaitanv, line 407: could we key this dedup on a `HashSet<(line, finding_type)>` instead of `findings.iter().any(...)` for every node in the recursive walk?

src/scanner/core.rs
```diff
@@ -313,7 +419,15 @@ impl ScanningLogic {
         let mut cursor = node.walk();
         if cursor.goto_first_child() {
             loop {
-                Self::scan_node_for_assignments(cursor.node(), source, filepath, assignment_rules, language_support, findings, processed_lines);
+                Self::scan_node_for_assignments(
```
juangaitanv, line 423: do the recursive AST walks (`scan_node_for_assignments`, `collect_all_relevant_nodes`) need an explicit depth guard, per the project convention?

src/scanner/core.rs
```diff
@@ -1330,7 +1688,12 @@ impl VulnerabilityScanner {
...
-    pub fn find_vulnerabilities_parallel(&self, root_dir: &str, language_name: &str, show_progress: bool) -> Result<Vec<Finding>> {
+    pub fn find_vulnerabilities_parallel(
+        &self,
+        root_dir: &str,
+        language_name: &str,
+        show_progress: bool,
+    ) -> Result<Vec<Finding>> {
         let files = self.discover_files(root_dir)?;
         if files.is_empty() {
             println!("No {} files found in {}", language_name, root_dir);
```
juangaitanv: should this `println!` check `show_progress`? json/csv output goes to the same stdout fusion parses.

src/scanner/core.rs
```diff
@@ -1613,14 +2017,18 @@ impl VulnerabilityScanner {
...
+
         if has_taint_rules && filtered_files.len() > 1 && !should_skip_cross_file {
             if show_progress {
                 log::info!("Performing cross-file taint analysis...");
             }
 
             let mut multi_file_analyzer = MultiFileTaintAnalyzer::new();
-            match multi_file_analyzer.analyze_cross_file_flows(&files_by_language, &taint_rules, language_filter) {
+            match multi_file_analyzer.analyze_cross_file_flows(
+                &files_by_language,
```
juangaitanv: should cross-file analysis receive `filtered_files` rather than the original `files_by_language`, so minified/test/doc/excluded files stay out?

src/scanner/core.rs
```diff
@@ -2527,52 +2959,61 @@ impl MultiFileTaintAnalyzer {
...
-            target_files.len(), language);
-        
-        // Initialize the new DataFlowTracer with the appropriate language files
+        log::debug!(
+            "[CROSS_FILE_NEW] Analyzing {} {} files for cross-file taint flows",
+            target_files.len(),
+            language
+        );
+
         let mut data_flow_tracer = DataFlowTracer::new();
```
juangaitanv, line 3034: does `DataFlowTracer` need to initialize an import/function resolver before cross-file analysis? it looks like it only holds caches now (`struct DataFlowTracer`), which may be what forces the hardcoded module fallback below.

src/scanner/core.rs
```diff
@@ -2527,52 +2959,61 @@ impl MultiFileTaintAnalyzer {
...
 
-        log::debug!("[CROSS_FILE_NEW] Analyzing {} files with sinks", self.file_imports.len());
+        log::debug!(
+            "[CROSS_FILE_NEW] Analyzing {} files with sinks",
+            self.file_imports.len()
+        );
 
         // For each file with sinks, use the new precise analysis
         for (sink_file, imports) in &self.file_imports {
             for sink_info in &imports.taint_sinks {
```
juangaitanv, line 3077: do we need to dedup cross-file findings before pushing per `sink_info`, the way the single-file path does via `mark_flow_processed`?

src/scanner/core.rs
```diff
@@ -2786,1108 +3245,217 @@ impl MultiFileTaintAnalyzer {
...
-                if !cursor.goto_next_sibling() {
-                    break;
+
+            // Check for taint sinks (eval, exec, os.system, etc.)
+            if let Some(sink_pattern) =
+                Self::extract_taint_sink_pattern(&node, source, rule_deduplicator)
+            {
+                // Extract variables from function call arguments
+                let used_variables = CommonUtils::extract_all_variables(&node_text);
+                if let Some(first_var) = used_variables.first() {
```
juangaitanv: does `.first()` of the sorted/deduped variables pick the intended argument for multi-arg sinks?

src/scanner/core.rs
```diff
@@ -4128,14 +3673,21 @@ impl DataFlowTracer {
...
-        log::debug!("[COMPUTE_VARIABLE_SOURCE] Analyzing variable '{}' in {}::{}", 
-            variable_name, file_path, function_name);
+        log::debug!(
+            "[COMPUTE_VARIABLE_SOURCE] Analyzing variable '{}' in {}::{}",
+            variable_name,
+            file_path,
+            function_name
+        );
 
         // Read the source code as text and do simple string analysis
```
juangaitanv, line 3799: should `compute_variable_source` scope the `{var} =` search to the function body rather than the whole file?

src/scanner/core.rs
```diff
@@ -4146,40 +3698,51 @@ impl DataFlowTracer {
             let line = line.trim();
             if line.starts_with(&format!("{} =", variable_name)) {
```
juangaitanv, line 3840: could we use `splitn(2, '=')` so `x = a == b` and augmented assignments like `x += taint` parse correctly?

src/scanner/core.rs
```diff
@@ -4198,69 +3763,13 @@ impl DataFlowTracer {
...
-                AnalysisResult::DefinitelySafe
-            },
-            AnalysisResult::Unknown { reason } => {
-                log::debug!("[TRACE_IMPORTED] Function '{}' analysis inconclusive: {}", import_info.local_name, reason);
-                AnalysisResult::Unknown { reason }
-            }
-        }
-    }
-
     fn trace_local_assignment_taint(
```
juangaitanv, line 3916: should `trace_local_assignment_taint` and `analyze_function_taint_behavior` track a visited set or depth limit before recursing into each other?

src/scanner/core.rs
```diff
@@ -4310,49 +3821,49 @@ impl DataFlowTracer {
...
                         // Create a cross-file taint flow from the original source to the current sink
                         let cross_file_flow = VerifiedTaintFlow {
                             source_file: flow.source_file,
                             source_function: flow.source_function,
                             source_pattern: flow.source_pattern,
                             source_line: flow.source_line,
-                            
+
                             sink_file: file_path.to_string(),
                             sink_function: function_name.to_string(),
```
juangaitanv, line 3992: does `sink_function` end up holding the callee name after `function_name` is reassigned to the called function above?

src/scanner/core.rs
```diff
@@ -4386,49 +3900,76 @@ impl DataFlowTracer {
...
+            calling_file
+        );
 
-        let calling_dir = std::path::Path::new(calling_file).parent()
+        let calling_dir = std::path::Path::new(calling_file)
+            .parent()
             .and_then(|p| p.to_str())
             .unwrap_or("");
 
         // Known patterns from the test files
```
juangaitanv, line 4079: what case do the `get_database_config`->`module_a.py` and `propagate_*`->`module_b.py` arms handle outside the test fixtures?

src/scanner/core.rs
```diff
@@ -4485,33 +4038,38 @@ impl DataFlowTracer {
...
+                        );
+
                         let nested_result = self.trace_local_assignment_taint(
-                            file_path, function_name, return_expr, line_num + 1,
-                            "function_return", line_num + 1, "return_value",
-                            rule_deduplicator
+                            file_path,
+                            function_name,
+                            return_expr,
+                            line_num + 1,
```
juangaitanv: should cross-file source line numbers be offset to absolute file lines rather than relative to the extracted body?

src/scanner/utils.rs
```diff
@@ -1,27 +1,25 @@
...
 
 /// Check if a file path matches a glob pattern (delegates to common utils)
 pub fn matches_glob_pattern(pattern: &str, file_path: &str) -> bool {
     crate::common::CommonUtils::matches_file_pattern(pattern, file_path)
 }
 
-
-
-
 pub fn is_git_ignored(path: &Path) -> bool {
```
juangaitanv, line 30: could we build the gitignore matcher once instead of walking the repo root per file? the cache is keyed per path, so each new file is a miss that triggers a full-repo walk.

src/scanner/utils.rs
```diff
@@ -206,7 +224,11 @@ pub fn discover_files_by_language_with_progress(root_dir: &str, parallel: bool,
 }
 
 /// Internal parallel file discovery implementation
-fn discover_files_parallel(root_dir: &str, estimated_languages: usize, show_progress: bool) -> Result<BTreeMap<String, Vec<PathBuf>>> {
+fn discover_files_parallel(
+    root_dir: &str,
+    estimated_languages: usize,
+    _show_progress: bool,
```
juangaitanv, line 287: is dropping the "discovered N files" note in the parallel path (`_show_progress`) intended? sequential mode still prints it.

src/scanner/prefilter.rs
```diff
@@ -74,11 +80,12 @@ impl PreFilter {
     /// Simple text/doc file detection
     fn is_text_or_doc_file(&self, file_path: &str) -> bool {
```
juangaitanv, line 83: should the text/doc skip run after the `is_malicious_scan` check, so malware scans still inspect `.svg` and binary files?

src/scanner/prefilter.rs
```diff
@@ -131,8 +139,11 @@ impl PreFilter {
     /// Simple heuristic to detect minified content (edge cases only)
     fn is_likely_minified_content(&self, file_path: &str) -> bool {
```
juangaitanv, line 142: nit: could `is_likely_minified_content` cap the read to the first N KB instead of reading the whole file?

src/main.rs
```diff
@@ -61,36 +71,43 @@ fn main() -> Result<()> {
...
 
+        // Combine findings
         simple_findings.append(&mut taint_findings);
         simple_findings
     };
 
+    // Output results
     let duration = start_time.elapsed();
 
     match cli.output_format.as_str() {
```
juangaitanv, line 106: should an unrecognized `--format` value error rather than fall through to text output? fusion parses this.

src/main.rs
```diff
@@ -61,36 +71,43 @@ fn main() -> Result<()> {
...
         "json" => print_findings_json(&findings),
```
juangaitanv: do we need to propagate write errors from `print_findings_json`/`print_findings_csv`, e.g. a broken pipe when fusion closes the reader?

src/models.rs
```diff
@@ -347,11 +373,17 @@ pub struct Cli {
     pub skip_minified: Option<bool>,
 
     /// Filter by code type (frontend, backend, or both)
-    #[arg(long, help = "Filter by code type: frontend, backend, or both (default: both)")]
+    #[arg(
+        long,
+        help = "Filter by code type: frontend, backend, or both (default: both)"
+    )]
     pub code_type: Option<String>,
```
juangaitanv, line 380: should an invalid `--code-type` error rather than silently disabling the filter? maybe a clap `value_parser`.

src/lib.rs
```diff
@@ -1,33 +1,58 @@
...
-pub mod code_type_detector;
 pub mod ui;
 
-// Re-export the main types and functions that main.rs needs
+pub use scanner::modes::run_taint_analysis_with_verbosity;
 pub use scanner::{
-    VulnerabilityScanner, ScanningLogic, PreFilter, FilterStats, 
-    run_explicit_scan, run_auto_detection_scan, run_taint_analysis
+    run_auto_detection_scan, run_explicit_scan, run_taint_analysis, FilterStats, PreFilter,
+    ScanningLogic, VulnerabilityScanner,
```
juangaitanv, line 44: do `PreFilter`, `ScanningLogic`, and `traverse_calls_only` need to be public while the API is pre-1.0, or could the exported surface stay narrower?

yhoztak (approved):
LGTM

src/scanner/core.rs
juangaitanv: Python is covered by the main `scan_file_with_rules` path, so the `scan_assignments` gate intentionally excludes it; the Python node-kind handling in `scan_node_for_assignments` is a documented guard for a future extension, not a dead path. Leaving as-is.

src/scanner/core.rs
juangaitanv: The `findings.iter().any(...)` uses a narrower `(line, finding_type)` key than `processed_lines`, so it can't just reuse that HashSet — it'd need a second one. Impact is small for per-file finding counts, so deferring as a cleanup.

src/scanner/core.rs
juangaitanv: These walk the tree-sitter CST, whose depth is bounded by source nesting and the parser rather than external input; the `max_depth` convention targets subprocess/agent recursion, and no other Sighthound AST traversal carries such a guard. Leaving as-is.

src/scanner/core.rs
juangaitanv: Fixed — the two unconditional `println!`s in `find_vulnerabilities_parallel` are now wrapped in `if show_progress`, so they no longer corrupt JSON/CSV stdout.

src/scanner/core.rs
juangaitanv: Fixed — cross-file analysis now receives a per-language map rebuilt from `filtered_files`, so minified/test/doc/excluded files no longer participate.

src/scanner/core.rs
juangaitanv: Fixed — `DataFlowTracer` now carries an import map populated from the imports already parsed in `build_import_export_maps`, and `find_function_source_file` consults it first. The hardcoded fixture fallback is kept for now (removal tracked separately).

src/scanner/core.rs
juangaitanv: Fixed — added a flow-identity dedup before pushing cross-file findings (keyed on sink file/line/variable/pattern + source file/line/pattern), mirroring the single-file `mark_flow_processed` while keeping genuinely distinct tainted variables as separate findings.

src/scanner/core.rs
juangaitanv: Fixed — sink discovery now records one `TaintSinkInfo` per used variable instead of the lexicographically-first one (`extract_all_variables` sorts/dedups), so a non-alphabetically-first tainted arg in a multi-arg sink is no longer dropped.

src/scanner/core.rs
juangaitanv: Fixed — the `{var} =` search is now scoped to the enclosing function body via `extract_function_body`, with absolute file line numbers preserved (falls back to the whole-file scan if the function can't be located).

src/scanner/core.rs
juangaitanv: Fixed — RHS extraction now uses `split_once('=')` so `x = a == b` and kwarg `=` keep the full RHS. Augmented assignments aren't matched by this `{var} =` path (noted in a comment).

src/scanner/core.rs
juangaitanv: Fixed — threaded a `BTreeSet<(file, function)>` visited set through both `trace_local_assignment_taint` and `analyze_function_taint_behavior`, returning `Unknown` on re-entry so a cyclic call graph can't recurse to stack overflow.

src/scanner/core.rs
juangaitanv: Fixed — renamed the shadowing inner binding to `callee_name`, so `sink_function` now resolves to the enclosing function that actually contains the sink.

src/scanner/core.rs
juangaitanv: They're hardcoded fixture shortcuts, kept harmless by the `exists()` check, so they're effectively dead in production. Removing them is worthwhile but deferring to a separate cleanup alongside the import-resolver work.

src/scanner/core.rs
juangaitanv: Fixed — `extract_function_body` now also returns the function's absolute start line, and source/sink lines are computed as `start_line + body_index`, so they map to real file lines instead of body-relative ones.

src/scanner/utils.rs
juangaitanv: Fixed — gitignore matching now builds one `Gitignore` matcher per repo root (walking the repo once) and reuses it per file, replacing the per-file full-repo walk. Note: matching is now strictly by `.gitignore` rules, so hidden files not listed there are now scanned (intentional per discussion).

src/scanner/utils.rs
juangaitanv: Both parallel and sequential discovery dropped this print together in this PR, so it's consistent rather than a parallel-only omission; if we want it back it should route through `crate::ui`. Leaving as-is.

src/scanner/prefilter.rs
juangaitanv: Fixed — moved the text/doc skip below the `is_malicious_scan` guard in both `should_scan_file` and `check_file_with_reason`, so malware scans now inspect `.svg`/binary files.

src/scanner/prefilter.rs
juangaitanv: Fixed — `is_likely_minified_content` now reads at most the first 8 KB; the `<=3 lines && >2000 bytes` decision is unchanged for all files (equivalent at the byte cap).

src/main.rs
juangaitanv: Valid — the `_` arm silently falls back to text and there's no clap guard on `output_format`. Low risk since fusion controls the flag, so deferring the `possible_values` hardening to a follow-up.

src/main.rs
juangaitanv: Fixed — `print_findings_json`/`print_findings_csv` now return `Result`, write through a `BufWriter`, and propagate I/O errors through `main`, so a closed reader exits cleanly instead of panicking.

src/models.rs
juangaitanv: Valid — an invalid `--code-type` becomes `None` and silently disables the filter. A clap `value_parser` limited to `frontend|backend|both` is the clean fix; deferring as a follow-up since fusion passes known-good values.

src/lib.rs
juangaitanv: `main.rs` reaches all three via qualified crate paths rather than these re-exports, so narrowing the surface is non-breaking but best deferred to a post-1.0 cleanup. Leaving as-is for now.

PR #23, opened 2026-06-15
Harden taint strictness and add boundary test suite

asadeddin:
## Summary

Taint strictness hardening and a new boundary test suite. Builds on #22.

**Merge order:** PR 2 of 2 — merge after #22. PRs #20 and #21 were closed.

---

### Safe-value classification (`src/scanner/core.rs`)
- **`classify_value_source()`** — decides whether an expression is a proven-safe literal, a taint source, or unknown
- **`is_safe_literal_expression()`** — string literals, safe static file reads
- **Environment variable awareness** — `extract_env_key()` distinguishes user-controlled keys (`USER_INPUT`, `REQUEST_DATA`) from config/secret keys (`API_KEY`, `DATABASE_URL`, `PORT`); `os.environ.setdefault` is config-safe unless its default value is itself tainted

### Sanitizer-aware assignments & sinks
- **`expression_has_sanitizer()` / `expression_has_any_sanitizer()`** — checks sanitizers in assignment RHS before recording taint (e.g. `shlex.quote(user)` suppresses flow)
- **Actionable sink filtering** — `is_actionable_sink_node()` limits sink matching to real call/invocation nodes
- **Bare-call source matching** — `is_bare_call_source_pattern()` / `matches_bare_call_source()` for patterns like `input(` without false-matching substrings

### PHP & variable handling
- **`extract_php_variables()`**, **`normalize_variable()`** — proper `$var` tracking in PHP taint flows

### File discovery & CLI (`src/scanner/modes.rs`, `src/scanner/utils.rs`, `src/models.rs`, `src/main.rs`)
- **`--include-test-fixtures`** — scans `tests/` / `test/` dirs normally skipped by prefilter (for fixture validation)
- New `*_with_options` scan APIs threading the flag through discovery and scanning
- **`deduplicate_findings()`** — collapses duplicate reports (same file, line, type, snippet)
- Show `--help` when invoked without a scan target

---

### `rules/javascript/frontend_taint_security.ron`
Tightened overly-broad patterns that caused false positives:
- **Removed sources:** `*.response`, `fetch(`, `XMLHttpRequest`, `axios.*`
- **Removed sinks:** `*.onload`, `*.onerror`, `addEventListener`, `Function(` (kept `new Function(`)
- **Narrowed to:** `*.responseText`, `*.response.data`, `*.responseJSON`

---

## New test suite: `tests/strictness/` (21 tests)

| Module | What it guards |
|--------|----------------|
| `cross_file_taint` | Phantom flows rejected; valid 1–3 file chains detected; config env vars safe |
| `taint_true_positives` | Cross-file accuracy fixtures report real vulns |
| `taint_true_negatives` | Safe cross-file sources stay silent |
| `sanitizer_and_scope` | `shlex.quote` suppresses findings; no scope leakage across functions |
| `search_conditions` | `subprocess` only flagged with `shell=True`; Java `Html.escape` suppresses XSS |
| `django_security` | Django fixture vulns detected; escaped `HttpResponse` not flagged |
| `false_positive_regressions` | Cleartext transmission fixture — TP section flagged, FP section silent |
| `language_coverage` | Go/Ruby/PHP/C# parity fixtures; `--include-test-fixtures` flag works |

Fixtures staged into temp dirs with neutral module names so the test-directory prefilter doesn't skip them.

### Test runner refactors
- `tests/test_files/accuracy_tests/run_accuracy_tests.py` — simplified
- `tests/test_files/multi_file_taint_tests/run_comprehensive_tests.py` — simplified

---

## Files to focus review on

| Priority | Files |
|----------|-------|
| High | `src/scanner/core.rs` (safe literals, sanitizers, env keys) |
| Medium | `src/main.rs`, `src/scanner/modes.rs`, `src/scanner/utils.rs`, `rules/javascript/frontend_taint_security.ron` |
| Tests | `tests/strictness/` (new), accuracy runner diffs |

## Test plan

- [x] `cargo test --test strictness_tests` (21 tests)
- [x] `cargo test` (82 tests total)
- [ ] `python3 tests/test_files/multi_file_taint_tests/run_comprehensive_tests.py`
- [ ] `python3 tests/test_files/accuracy_tests/run_accuracy_tests.py`

---
**Integration branch:** Merges into [`feature/more_language_support`](https://github.com/Corgea/Sighthound/tree/feature/more_language_support) — **not** `main`.

**Merge order:** #16–#19 (merged) → #22 → **#23**

### DIFF
--- a/rules/javascript/frontend_taint_security.ron
+++ b/rules/javascript/frontend_taint_security.ron
@@ -61,9 +61,8 @@
                 "e.data",
                 "message.data",
                 
-                // Network responses (external data)
+                // Network response text that may be rendered as HTML.
                 "*.responseText",
-                "*.response",
                 "fetch(*).then",
                 "XMLHttpRequest.responseText",
                 "axios.get",
@@ -92,10 +91,8 @@
                 "iframe.src",
                 "*.srcdoc",  // iframe srcdoc
                 
-                // Attribute manipulation
+                // Attribute manipulation that can inject executable markup.
                 "*.onclick",
-                "*.onload",
-                "*.onerror",
                 "*.setAttribute",
                 // Code execution
                 "eval(",
@@ -276,9 +273,8 @@
                 // PostMessage (external data)
                 "event.data",
                 
-                // Network responses (external data)
+                // Network response text that may be rendered as HTML.
                 "*.responseText",
-                "*.response",
                 "fetch(*).then",
                 "XMLHttpRequest.responseText",
                 "axios.*.then",
@@ -308,11 +304,8 @@
                 "*.formAction",
                 "*.setAttribute",
                 
-                // Event handlers
+                // Event handlers that execute attacker-controlled code.
                 "*.onclick",
-                "*.onload",
-                "*.onerror",
-                "addEventListener",
                 
                 // Template/rendering (dangerous)
                 "template.innerHTML",
@@ -1777,12 +1770,10 @@
                 "prompt(",
                 "confirm(",
                 
-                // Network responses
+                // Network response values that can carry executable code.
                 "*.responseText",
-                "*.response",
-                "fetch(",
-                "XMLHttpRequest",
-                "axios.*",
+                "*.response.data",
+                "*.responseJSON",
                 
                 // Storage (could contain user data)
                 "localStorage.getItem",
@@ -1797,7 +1788,6 @@
             sinks: Some([
                 // Direct code execution
                 "eval(",
-                "Function(",
                 "new Function(",
                 
                 // Timed execution with string
--- a/rules/javascript/frontend_taint_security.ron
+++ b/rules/javascript/frontend_taint_security.ron
@@ -93,6 +93,8 @@
                 
                 // Attribute manipulation that can inject executable markup.
                 "*.onclick",
+                "*.onload",
+                "*.onerror",
                 "*.setAttribute",
                 // Code execution
                 "eval(",
@@ -148,7 +150,7 @@
                 // Exclude safe operations that commonly cause false positives
                 (
                     field: "pattern",
-                    operator: "not_contains", 
+                    operator: "not_contains",
                     value: "addEventListener",
                     condition_type: Some("exclude_safe_operations")
                 ),
@@ -306,7 +308,10 @@
                 
                 // Event handlers that execute attacker-controlled code.
                 "*.onclick",
-                
+                "*.onload",
+                "*.onerror",
+                "addEventListener",
+
                 // Template/rendering (dangerous)
                 "template.innerHTML",
                 "*.render",
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,5 +1,5 @@
 use anyhow::Result;
-use clap::Parser;
+use clap::{CommandFactory, Parser};
 use sighthound::scanner::core::{print_findings_csv, print_findings_json, print_findings_text};
 use sighthound::{
     run_auto_detection_scan, run_explicit_scan, run_taint_analysis,
@@ -23,12 +23,15 @@ fn main() -> Result<()> {
         return Ok(());
     }
 
-    // Check if root_dir is provided (required for actual scanning)
-    let root_dir = cli.root_dir.as_ref().ok_or_else(|| {
-        anyhow::anyhow!(
-            "Root directory is required for scanning. Use --help for usage information."
-        )
-    })?;
+    // Show help when invoked without a scan target.
+    let root_dir = match cli.root_dir.as_ref() {
+        Some(root_dir) => root_dir,
+        None => {
+            Cli::command().print_help()?;
+            println!();
+            return Ok(());
+        }
+    };
 
     // Initialize logger (respect RUST_LOG or --verbose flag)
     env_logger::Builder::from_env(
@@ -97,6 +100,7 @@ fn main() -> Result<()> {
 
         // Combine findings
         simple_findings.append(&mut taint_findings);
+        deduplicate_findings(&mut simple_findings);
         simple_findings
     };
 
@@ -110,4 +114,17 @@ fn main() -> Result<()> {
     }
 
     Ok(())
+}
+
+fn deduplicate_findings(findings: &mut Vec<sighthound::Finding>) {
+    let mut seen = std::collections::BTreeSet::new();
+    findings.retain(|finding| {
+        seen.insert((
+            finding.file.clone(),
+            finding.line,
+            finding.end_line,
+            finding.finding_type.clone(),
+            finding.snippet.clone(),
+        ))
+    });
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -73,7 +73,7 @@ fn main() -> Result<()> {
         ))?;
 
     // Handle all vulnerability scanning modes with unified flow
-    let findings = if cli.taint_analysis {
+    let mut findings = if cli.taint_analysis {
         // Only taint analysis
         run_taint_analysis(&cli, root_dir, show_progress)?
     } else if cli.simple_analysis {
@@ -100,9 +100,11 @@ fn main() -> Result<()> {
 
         // Combine findings
         simple_findings.append(&mut taint_findings);
-        deduplicate_findings(&mut simple_findings);
         simple_findings
     };
+
+    // Deduplicate findings across all analysis modes (keeps first occurrence order)
+    deduplicate_findings(&mut findings);
 
     // Output results
     let duration = start_time.elapsed();
@@ -125,6 +127,8 @@ fn deduplicate_findings(findings: &mut Vec<sighthound::Finding>) {
             finding.end_line,
             finding.finding_type.clone(),
             finding.snippet.clone(),
+            finding.severity.clone(),
+            finding.description.clone(),
         ))
     });
 }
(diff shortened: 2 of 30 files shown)


juangaitanv (commented):
Reviewed the taint-strictness changes (built + ran the suite locally: 83 tests pass). A few questions below, grouped inline. The one I'd most want resolved is the dotted-`input(` source change in `core.rs`, which dropped a flow the base branch still catches.

src/scanner/core.rs
```diff
@@ -94,37 +102,183 @@ impl TaintRuleDeduplicator {
...
 
+    fn is_bare_call_source_pattern(pattern: &str) -> bool {
+        let Some(name) = pattern.strip_suffix('(') else {
+            return false;
+        };
+
+        !name.is_empty() && name.chars().all(|c| c.is_ascii_alphanumeric() || c == '_')
+    }
+
+    fn matches_bare_call_source(pattern: &str, text: &str) -> bool {
```
juangaitanv, line 141: should `builtins.input(` / `six.moves.input(` still register as sources? the `.` prefix check now excludes them, and a `builtins.input()` flow the base branch flagged is no longer reported.

src/scanner/core.rs
```diff
@@ -4198,67 +4078,180 @@ impl DataFlowTracer {
...
+        let literal_concat = expr
+            .split('+')
+            .map(str::trim)
+            .filter(|part| !part.is_empty())
+            .all(Self::is_string_literal);
+
+        literal_concat && expr.contains('+')
+    }
+
+    fn is_string_literal(expression: &str) -> bool {
```
juangaitanv, line 4339: should this reject expressions like `"x" + tainted + "y"` that start and end with a quote, given the per-operand concat check below never runs once this returns true for the whole string?

src/scanner/core.rs
```diff
@@ -4198,67 +4078,180 @@ impl DataFlowTracer {
...
+        let expr = expression.trim();
+        (expr.len() >= 2 && expr.starts_with('"') && expr.ends_with('"'))
+            || (expr.len() >= 2 && expr.starts_with('\'') && expr.ends_with('\''))
+    }
+
+    fn is_safe_static_file_read(expression: &str) -> bool {
```
juangaitanv, line 4368: should the static-file safelist require the whole path argument to be one literal rather than the first quoted prefix, e.g. `open("config/" + user_input)`?

src/scanner/core.rs
```diff
@@ -4198,67 +4078,180 @@ impl DataFlowTracer {
...
-                AnalysisResult::DefinitelySafe
-            },
-            AnalysisResult::Unknown { reason } => {
-                log::debug!("[TRACE_IMPORTED] Function '{}' analysis inconclusive: {}", import_info.local_name, reason);
-                AnalysisResult::Unknown { reason }
+        if self.is_safe_literal_expression(expr) {
+            return ValueSourceClassification::Safe("literal expression".to_string());
+        }
+
+        if expr.contains("setdefault(") {
```
juangaitanv: does `setdefault(` need to stay safe when its default argument carries taint, given it returns that value?

src/scanner/core.rs
```diff
@@ -4198,67 +4078,180 @@ impl DataFlowTracer {
...
+                if let Some(end) = rest.find(quote) {
+                    return Some(rest[..end].to_ascii_uppercase());
+                }
             }
         }
+
+        None
+    }
+
+    fn is_user_controlled_env_key(key: &str) -> bool {
```
juangaitanv, line 4434: should `API_KEY` match here? the PR summary lists it as user-controlled but no branch covers it.

src/main.rs
```diff
@@ -61,36 +74,57 @@ fn main() -> Result<()> {
...
-        "text" | _ => print_findings_text(&findings, cli.verbose, cli.summary_only, duration),
+        _ => print_findings_text(&findings, cli.verbose, cli.summary_only, duration),
     }
 
     Ok(())
-} 
\ No newline at end of file
+}
+
+fn deduplicate_findings(findings: &mut Vec<sighthound::Finding>) {
```
juangaitanv, line 121: should dedup also run in `--taint-analysis` and `--simple-analysis` modes, or is overlap only expected in the combined path?

src/main.rs
```diff
@@ -61,36 +74,57 @@ fn main() -> Result<()> {
...
+    let mut seen = std::collections::BTreeSet::new();
+    findings.retain(|finding| {
+        seen.insert((
+            finding.file.clone(),
+            finding.line,
```
juangaitanv, line 126: do we need `severity` or the rule message in the dedup key, or are two findings sharing file/line/end_line/type/snippet always the same finding?

rules/javascript/frontend_taint_security.ron
```diff
@@ -92,10 +91,8 @@
                 "iframe.src",
                 "*.srcdoc",  // iframe srcdoc
                 
-                // Attribute manipulation
+                // Attribute manipulation that can inject executable markup.
```
juangaitanv, line 94: are `*.onload` / `*.onerror` / `addEventListener` intentionally dropped as sinks, e.g. `img.onerror = location.hash`?

rules/javascript/frontend_taint_security.ron
```diff
@@ -1797,7 +1788,6 @@
             sinks: Some([
                 // Direct code execution
                 "eval(",
-                "Function(",
                 "new Function(",
```
juangaitanv, line 1796: is dropping bare `Function(` while keeping `new Function(` intended, leaving direct `Function(userInput)()` uncovered?

src/scanner/utils.rs
```diff
@@ -149,71 +163,114 @@ pub fn detect_language_from_path(file_path: &Path) -> Option<&'static str> {
...
-pub fn discover_files_by_language_with_progress(root_dir: &str, parallel: bool, show_progress: bool) -> Result<BTreeMap<String, Vec<PathBuf>>> {
+pub fn discover_files_by_language_with_progress(
+    root_dir: &str,
+    parallel: bool,
+    show_progress: bool,
+) -> Result<BTreeMap<String, Vec<PathBuf>>> {
+    discover_files_by_language_with_progress_and_options(root_dir, parallel, show_progress, false)
+}
+
+pub fn discover_files_by_language_with_progress_and_options(
```
juangaitanv, line 290: could the parallel `*_with_options`, progress, and sequential discovery wrappers collapse into one options struct, to reduce the near-duplicate surface?

src/scanner/core.rs
juangaitanv: Fixed — matches_bare_call_source now carves out the `builtins.` / `six.moves.` qualifiers, so `builtins.input(` and `six.moves.input(` register as sources again while `obj.input()` stays excluded.

src/scanner/core.rs
juangaitanv: Fixed — is_string_literal now matches only a single atomic quoted literal (it scans for the matching closing quote), so `"x" + tainted + "y"` falls through to the per-operand concat check instead of being treated as a safe literal.

src/scanner/core.rs
juangaitanv: Fixed — is_safe_static_file_read now requires the whole path argument to be a single literal (the char after the closing quote must be `)` or `,`), so `open("config/" + user_input)` is no longer safelisted.

src/scanner/core.rs
juangaitanv: Fixed — dropped the blanket `setdefault(` safe-guard so a tainted default (the returned value) is classified and `os.environ.setdefault` reaches the env-key path; config-key setdefaults still classify as safe.

src/scanner/core.rs
juangaitanv: Resolved as config, not a source — `API_KEY` / `*_API_KEY` are now in is_config_env_key (a secret/config value, not user-controlled input), and the PR description has been corrected to match.

src/main.rs
juangaitanv: Fixed — deduplicate_findings is now called once after the mode dispatch, so `--taint-analysis` and `--simple-analysis` dedup too, not just the combined path.

src/main.rs
juangaitanv: Fixed — added severity and description to the dedup key, so two distinct findings that share file/line/type/snippet but differ in severity or rule message are no longer collapsed.

rules/javascript/frontend_taint_security.ron
juangaitanv: Fixed — restored `*.onload` / `*.onerror` to both DOM-XSS sink lists and re-added `addEventListener` to the basic-XSS sinks; the PR's other intentional narrowing (`Function(`, response sources) is left intact.

rules/javascript/frontend_taint_security.ron
juangaitanv: Bare `Function(` is still a sink in four other rules (lines ~99, 816, 883, 1384); the removal was scoped to the one rule whose sources were narrowed to structured response shapes, so `Function(userInput)()` stays covered — no gap.

src/scanner/utils.rs
juangaitanv: The only real duplication is four one-liner legacy wrappers, and modes.rs already calls the options function directly — worth a separate cleanup issue rather than this PR.

PR #26, opened 2026-06-26
Repair test harnesses against the unified rule API (cargo test green)

juangaitanv:
## Repair test harnesses against the unified rule API (`cargo test` green)

Makes `cargo test` green again by fully realigning all three harnesses to the current unified-rule API and re-enabling the modules that #12 quarantined. **No `src/` changes** — tests only.

### Background
The test harnesses were stale against the scanner's refactor to the unified RON rule model. #12 (`fix-test-crate-rename`) renamed the `find_vulns` crate references to `sighthound`, but to get a compile it **quarantined five modules** — `file_pattern_tests`, `directory_loading_tests`, `rule_deserialization_tests`, `pattern_matching_tests`, and the **entire `integration_tests` harness** — and left `end_to_end_injection_tests` asserting pre-refactor behavior. As a result `cargo test` on `main` still fails (4 of 5 end_to_end tests fail at runtime) and runs with significantly reduced coverage.

### What this PR does
- Replaces the removed categorized `Rules` fields (`malware_detection`, `injection_sinks`, `crypto_rules`, `path_traversal`, `other`) with the unified accessors (`get_search_rules` / `get_taint_rules` / `count_rules` / `get_rules_by_category`).
- `rule_matches_pattern` → `rule_matches_pattern_unified`; `validate_rule_patterns` → `validate_unified_rule_patterns`; `Rule` → `UnifiedRule`; `scanner::shared::ScanningLogic` → `scanner::ScanningLogic`.
- Converts RON fixtures to the unified `(rules: [...])` schema.
- Adapts assertions to current scanner behavior — each documented with an in-file `// note:` comment. Examples: injection gating now keys on `category: "injection"` + `has_injection_pattern` (so the e2e SQL/command samples use tainted-concat forms the gating actually fires on; literal-string sinks remain correctly unflagged), `match_pattern` is substring-based, and quoted strings classify as literal nodes. No assertion was weakened to a no-op.
- Re-enables all quarantined modules in `tests/unit/main.rs` and `tests/integration/main.rs`.

### Result
```
cargo test
  unit_tests        60 passed
  integration_tests  7 passed
  end_to_end_tests   5 passed
  = 72 passed, 0 failed, 0 ignored
```

### Follow-up
Once this lands, re-add `- run: cargo test` to `.github/workflows/ci.yml` (the OSS-readiness PR left a TODO there and gates on `cargo build --release` only until this is green).


tests/unit/injection_pattern_tests.rs
```diff
@@ -1,8 +1,16 @@
 use sighthound::language::{get_language_support, LanguageSupport};
 use sighthound::rules::{check_for_injection_pattern, is_literal_node};
-use sighthound::scanner::core::ScanningLogic;
+use sighthound::scanner::ScanningLogic;
 use sighthound::parser::{LanguageParser, get_node_text};
 
+// note: in the unified-scanner refactor, `check_for_injection_pattern` became a
```
leenk7991, line 6: should we also add an e2e positive test coverage (via taint analysis)

tests/unit/django_xss_prevention_tests.rs
```diff
@@ -59,55 +56,37 @@ mod django_xss_tests {
...
+                for rule in &rules.rules {
+                    assert!(rule.pattern.is_some() || rule.patterns.is_some(),
+                           "Each rule should have either pattern or patterns");
                 }
             },
             Err(e) => {
                 println!("Warning: Failed to load XSS prevention rules: {}", e);
-                // This is acceptable for now - the rules file may have syntax issues
-                // but we don't want the test to fail completely
+                // This is acceptable - the rules file may be absent in this layout.
```
leenk7991: won't this make the test always pass? is there a away we can just assert missing file error handling?

tests/unit/pattern_matching_tests.rs
```diff
@@ -71,194 +100,98 @@ mod pattern_matching_tests {
...
+        let valid_regex = make_rule(Some("regex:^[a-z]+$"), None);
+        assert!(validate_unified_rule_patterns(&valid_regex).is_ok());
+
+        // Invalid regex in the single `pattern` field
+        let invalid_regex = make_rule(Some("regex:[unclosed"), None);
+        assert!(validate_unified_rule_patterns(&invalid_regex).is_err());
+
+        // Invalid regex inside the `patterns` list
+        let invalid_regex_in_list = make_rule(None, Some(vec!["ok", "regex:(("]));
+        assert!(validate_unified_rule_patterns(&invalid_regex_in_list).is_err());
```
leenk7991, line 186: should we add a test case for `make_rule(None, None)`? and other empty string/empty patterns list combination

tests/unit/django_xss_prevention_tests.rs
juangaitanv: Done — replaced the always-pass match: the test now asserts the absent xss_prevention.ron returns Err and checks rule structure against the shipped general_security.ron (search-mode rules; taint rules carry sources/sinks). Pushed in 50c2ed5.

tests/unit/pattern_matching_tests.rs
juangaitanv: Done — added make_rule(None, None), empty-string, and empty-patterns-list cases asserting .is_ok(): the unified validator only checks regex compilation and no longer rejects contentless search rules, so these document that permissive contract against regressions. Tightening the validator itself would be a separate change. Pushed in 50c2ed5.

tests/unit/injection_pattern_tests.rs
juangaitanv: Positive e2e injection detection already runs in tests/end_to_end/end_to_end_injection_tests.rs through the full scanner pipeline (Python/Java/JS). What is missing is a dedicated taint-flow (source->sink) positive case; since that is new coverage beyond this harness-repair PR, I will file a follow-up rather than fold it in here.

leenk7991 (approved):

PR #27, opened 2026-06-30
docs: correct outdated claims across Sighthound READMEs

juangaitanv:
## What

Surgical corrections to outdated/incorrect claims across the three Sighthound README files. Every change was verified against the actual code, `Cargo.toml`/`Cargo.lock`, CI config, and git history — no speculative edits.

## Why / changes

**README.md**
- **Rust `1.70+` → `1.85+`** (badge + prerequisites): the `harness` bin is `edition = "2024"`, which `cargo build --release` builds, so the real floor is Rust 1.85 (Dockerfile pins `rust:1.89`).
- **Broken `cargo run -- …` examples fixed → `cargo run --bin sighthound -- …`**: the crate now has two bins (`sighthound` + `harness`) and no `default-run`, so bare `cargo run` errors with "could not determine which binary to run." Verified the corrected form runs.
- **TSX language row**: `— (parser only)` → `✅ (shares JavaScript rules)` — `rules.rs` `load_embedded_rules` serves the JS embedded frontend+taint rules for `tsx`.
- **CLI options block**: documented the real but previously-missing flags (`--simple-analysis`, `--code-type`, `--language-filter`, `--rules-dir`, `--use-file-rules`, `--version`); corrected `--skip-minified` to take a `<BOOL>` value and noted `--rules-dir` pairs with `--use-file-rules`; added the "no flag → runs both" note.
- **Removed `Incremental Analysis: Cache results for unchanged files`** from Optimization Features — no such code exists; it's already listed as a roadmap TODO (self-contradiction).
- **Dropped the "tests under repair / not yet green" caveats** — the test-harness-repair work landed and CI runs the full `make ci` pipeline (which includes tests).
- **Architecture tree**: added the real `cli.rs` and `code_type_detector.rs` modules.
- **Docker export note**: corrected to `./sighthound_release/` and noted `rules/` is exported alongside the binary.

**tests/README.md** — repointed fixtures to `tests/test_files/`; `test_fixtures/`, `test_scripts/`, and `test_data/` do not exist.

**tests/test_files/multi_file_taint_tests/README.md** — removed references to `TAINT_TESTING_STRATEGY.md`, which is not on disk.

## Verification
- Cross-checked each claim against `src/models.rs` (clap CLI), `src/rules.rs`, `Cargo.toml`, `Cargo.lock`, `.github/workflows/ci.yml`, and the on-disk tree.
- Confirmed `cargo run --bin sighthound -- --version` works (`sighthound 0.1.1`).
- Docs-only change; no source touched.

## Deliberately left (not doc-rot, flagged for follow-up)
- Performance/accuracy benchmarks (unverifiable, but not stale).
- "Built-in Rule Categories" list is non-exhaustive vs. the full rule set — expanding it is a separate task.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 A high-performance vulnerability scanner for source code using tree-sitter parsing and advanced taint flow analysis.
 
-[![Rust](https://img.shields.io/badge/rust-1.70+-orange.svg)](https://www.rust-lang.org)
+[![Rust](https://img.shields.io/badge/rust-1.85+-orange.svg)](https://www.rust-lang.org)
 [![License](https://img.shields.io/badge/license-MIT-blue.svg)](LICENSE)
 [![CI](https://github.com/Corgea/Sighthound/actions/workflows/ci.yml/badge.svg)](https://github.com/Corgea/Sighthound/actions/workflows/ci.yml)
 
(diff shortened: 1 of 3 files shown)


asadeddin (approved):
LGTM

PR #29, opened 2026-07-02
build: trim syntect features to drop vulnerable quick-xml

juangaitanv:
## Problem

Main's CI has been red since 2026-06-29: the `Dep audit` gate (`cargo audit`) fails on two high-severity (CVSS 7.5) advisories against `quick-xml 0.37.5`, both patched only in `>=0.41.0`:

- [RUSTSEC-2026-0194](https://rustsec.org/advisories/RUSTSEC-2026-0194) — quadratic run time when checking a start tag for duplicate attribute names (DoS)
- [RUSTSEC-2026-0195](https://rustsec.org/advisories/RUSTSEC-2026-0195) — unbounded namespace-declaration allocation in `NsReader` (memory-exhaustion DoS)

`quick-xml` enters the dependency tree solely through syntect's default `plist-load` feature (`syntect -> plist -> quick-xml`), and no `plist` release permits `quick-xml 0.41` (latest `plist 1.9.0` pins `^0.39.2`). Sighthound never uses `plist-load`: its only syntect usage is `SyntaxSet::load_defaults_newlines()`, `ThemeSet::load_defaults()`, and `HighlightLines` (`src/scanner/core.rs`).

## Fix

Disable syntect default features, keeping `default-syntaxes`, `default-themes`, and `regex-onig`. `plist` and `quick-xml` leave the dependency graph entirely.

- No dependency versions bumped
- No CI gate weakened, skipped, or ignored
- Unbreaks main's CI (red since the advisories published 2026-06-29)

## Verification

Local `make ci` (the exact command the `harness ci` job runs) on this branch: exit 0.

```
✓ Clippy (strict)
✓ Format check
✓ Dep audit
✓ Tests (96 passed)
✓ Coverage
✓ Arch: no orphan files
```

(Complexity/CRAP marks are advisory and pre-existing.) `make audit` also green.

PR #28, opened 2026-07-02
docs: language support and fusion-benchmarks results

asadeddin:
## Summary
- Expands the language support section with file extensions, what ships on `main` vs `feature/more_language_support`, and explicit gaps (Razor, C/C++)
- Replaces vague performance claims with measured fusion-benchmarks results (Sighthound vs Semgrep) including per-language recall
- Updates known limitations and roadmap to reflect current scanner coverage

## Test plan
- [x] Reviewed README rendering for table alignment and link targets
- [x] Verified benchmark figures match `fusion-benchmarks/bench_out/summary.json` from July 2026 run

Made with [Cursor](https://cursor.com)

PR #30, opened 2026-07-02
feat: multi-language support (PHP, C#, Go, Ruby, Java, HTML rule packs)

juangaitanv:
## What this branch adds

Multi-language scanning support, integrated via internal PRs #16–#19:

- **Grammar foundation**: tree-sitter grammars for PHP, C#, Go, Ruby (plus a bump to tree-sitter 0.26) wired in as Cargo features, all on by default.
- **LanguageSupport impls** and **extension mappings** for the new languages in `src/language.rs` / scanner glue (HTML attribute resolution, Python assignment scanning).
- **Rule packs per language** under `rules/<lang>/`: PHP, C# (incl. the HttpClient SSRF rule), Go, Ruby, Java, and HTML (Thymeleaf + XSS), plus refreshed Python and JavaScript packs (Node backend coverage).
- **Strictness test suite** (`tests/strictness/`) with per-language fixtures covering taint true positives/negatives, sanitizer respect, cross-file taint, and PHP/Go/Ruby/C# language coverage.

This makes the language support table added to the README in #28 accurate — those claims now match the shipped scanner.

## Merge of main

`main` was merged in (commit 1881369) to land this work on the repaired test harness (#26) — pre-existing unit/integration/e2e tests take main's repaired versions (real-behavior assertions, unified rule API), adapted to tree-sitter 0.26's `u32` child indexing. `src/` keeps this branch's multi-language additions; main's clippy/rustfmt baseline is preserved (both sides had converged on the same lint fixes independently).

**Note**: the Dep-audit gate failed on `quick-xml 0.37.5` (RUSTSEC-2026-0194/0195, pulled in via syntect's `plist-load` default feature). Applied the same syntect feature trim as open PR #29 (`default-features = false`, keep `default-syntaxes`/`default-themes`/`regex-onig`) so this branch's CI is green; it dedupes cleanly when #29 lands.

## Local verification (`make ci`)

| Gate | Result |
|---|---|
| Clippy (strict, `-D warnings`) | ✅ |
| Format check | ✅ |
| Dep audit | ✅ (after syntect trim) |
| Complexity (lizard) | ⚠️ advisory, non-blocking |
| Tests | ✅ 118 passed |
| Acceptance | ⚠️ skipped (no `.feature` files) |
| Coverage (run + LCOV + threshold) | ✅ |
| CRAP | ⚠️ advisory, non-blocking |
| Arch (no orphan files) | ✅ |

`make ci` exits 0. Smoke-tested the built binary against `tests/test_files/strictness_languages/` — findings produced for C#, Go, PHP, and Ruby fixtures.


src/rules.rs
```diff
@@ -10,24 +10,96 @@ use crate::language::LanguageSupport;
...
-const EMBEDDED_JAVASCRIPT_TAINT_RULES: &str =
-    include_str!("../rules/javascript/frontend_taint_security.ron");
-
-// Add more embedded rules as needed
-// const EMBEDDED_JAVA_RULES: &str = include_str!("../rules/java/security.ron");
-// const EMBEDDED_HTML_RULES: &str = include_str!("../rules/html/security.ron");
+// Embedded rules - rule files included at compile time, grouped per language.
+// Each entry must be a valid `Rules` RON document; exclusion-pattern files use a
+// different schema and are intentionally excluded.
+const EMBEDDED_PYTHON: &[&str] = &[
```
asadeddin: @juangaitanv rather than this structure, should we make it more dynamic? For example, fetch all rules in a directory rather than spelling them out 1 by 1?

rules/java/broken_auth.ron
```diff
@@ -0,0 +1,45 @@
+(
+    rules: [
+        // CWE-256: Plaintext storage of a password — setPassword called with a
+        // raw value (no hashing/PasswordEncoder).
+        (
+            id: Some("java-plaintext-password-storage"),
```
asadeddin, line 8: Might be high false positives. We should test.

rules/javascript/backend_node_security.ron
```diff
@@ -0,0 +1,371 @@
+(
```
asadeddin: We should probably split these out into category ron files like the other languages

rules/python/general_security.ron
```diff
@@ -1,50 +1,61 @@
 (
     rules: [
-        // Code injection vulnerabilities
+        // ---- Code injection (eval/exec) ----
```
asadeddin, line 3: Should we split this out more?

rules/python/working.ron
```diff
@@ -1,105 +1,7 @@
...
-            tags: Some(["pickle", "deserialization", "unsafe"]),
-            message: None,
-        ),
-    ]
-) 
\ No newline at end of file
+    // Intentionally empty: coverage consolidated into general_security.ron,
+    // command_injection.ron and cryptography.ron. The previous bare "eval"/"exec"
+    // substring patterns matched `cursor.execute(...)` and produced large numbers
+    // of CWE-95 false positives.
```
asadeddin: We should get rid of this file

rules/python/simple_test.ron
```diff
@@ -1,29 +1,4 @@
...
-            sanitizers: None,
-            conditions: None,
-            tags: None,
-            message: None,
-        )
-    ]
-) 
\ No newline at end of file
+    // Intentionally empty: eval() coverage lives in general_security.ron.
+    rules: []
```
asadeddin: We should get rid of this file

src/scanner/core.rs
```diff
@@ -3,24 +3,17 @@
 //! This module provides the main vulnerability scanning functionality including:
```
asadeddin: this file has become a god file and we should probably split it up in another PR. WDYT?

asadeddin (changes requested):
Some feedback

src/rules.rs
juangaitanv: Done — replaced the per-file lists with include_dir!-embedded language directories that load every .ron at startup (skipping exclusion_patterns.ron), so new rule files auto-wire; this had already silently missed simple_test.ron.

rules/java/broken_auth.ron
juangaitanv: Done — added an argument_not_sanitized condition (*encode*/*hash*/*crypt*) to the setPassword rule mirroring java/xss.ron so the PasswordEncoder idiom no longer fires, with a new strictness test proving safe-vs-vulnerable discrimination.

rules/javascript/backend_node_security.ron
juangaitanv: Done — split into per-category backend_*.ron files along the existing CWE section boundaries, matching the other language packs; the new include_dir loader auto-wires them.

rules/python/general_security.ron
juangaitanv: Done — moved deserialization, SSRF, XXE, XSS, and open-redirect rules into their own category files, leaving general_security.ron with the code-injection core (eval/exec, SSTI, dynamic import).

rules/python/working.ron
juangaitanv: Done — deleted working.ron and dropped its include; its coverage already lives in general_security/command_injection/cryptography.

rules/python/simple_test.ron
juangaitanv: Done — deleted; it was empty and never embedded or referenced by the loader.

src/scanner/core.rs
juangaitanv: Agreed — at ~5.8k lines it needs splitting (rule matching / taint tracking / traversal are natural seams), but doing it here would swamp this language-pack diff; filing a follow-up ticket.

asadeddin (approved):
LGTM

PR #33, opened 2026-07-10
Migrate crate to Rust 2024 edition

juangaitanv:
## Summary

- migrate the crate and rustfmt configuration to Rust 2024
- let the harness binary inherit the package edition
- apply formatter and Clippy rewrites across scanner code and tests

## Validation

- `git diff --check HEAD`
- `cargo fmt --all -- --check`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test --all-features`

## Before merge

- [ ] Resolve the MSRV mismatch: `Cargo.toml` and the README promise Rust 1.85, but the new `&& let` chains require Rust 1.88 or newer.


Cargo.toml
```diff
@@ -1,7 +1,8 @@
 [package]
 name = "sighthound"
 version = "0.1.1"
-edition = "2021"
+edition = "2024"
+rust-version = "1.88"
```
yhoztak, line 5: It's declared 1.88 here, but `.github/workflows/ci.yml` has ` dtolnay/rust-toolchain@stable`. Do we want to consolidate them?

yhoztak (approved):
LGTM
with minor comment

Ibrahimrahhal (approved):

Cargo.toml
juangaitanv: Done — kept the full suite on stable and added a Rust 1.88 all-target check that enforces the declared MSRV.

PR #37, opened 2026-07-10
output: braille summary visuals for scan Summary (TTY only)

buttonmonkey:
## What

Two small, dependency-free visual touches to the scan **Summary**, both **TTY-only**:

1. **Findings heat map** — the "Most affected files" list gains a per-file braille bar: **length** = finding count (normalized to the busiest file), **colour** = worst severity in that file (reusing `ui::severity_code`).
2. **Severity mix bar** — a single stacked braille bar under the `● N critical  ● N high  …` line, segments sized by each severity's share and coloured by severity.

## What it looks like

> Colour doesn't render in GitHub code blocks — run it in a terminal to see it. Below, bar **length** = count and bar **colour** = severity (critical = bold red, high = red, medium = yellow, low = green).

A real multi-file scan — heat map (per file) + severity mix bar (overall blend):

```
Summary
  ● 2 critical   ● 7 high   ● 1 medium   ● 1 low
  ⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿

     7  Command Injection
     2  Unsafe Deserialization
     1  Dynamic Import
     1  Open Redirect

Most affected files
  ⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿     4  ./app/api/handlers.py
  ⣿⣿⣿⣿⣿⣿⣿⡇⠀⠀     3  ./auth/session.py
  ⣿⣿⣿⣿⣿⠀⠀⠀⠀⠀     2  ./db/query.py
  ⣿⣿⡇⠀⠀⠀⠀⠀⠀⠀     1  ./utils/format.py
  ⣿⣿⡇⠀⠀⠀⠀⠀⠀⠀     1  ./web/redirect.js
```

A balanced severity mix — the stacked bar reads the blend at a glance:

```
Summary
  ● 5 critical   ● 5 high   ● 5 medium   ● 5 low
  ⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿

     5  Command Injection
     5  Dynamic Import
     5  Open Redirect
     5  Unsafe Deserialization
```

The mix bar is normalized to the total, so it shows the **blend** of severities rather than the volume.

## Pipe-safety (the important guarantee)

Both visuals are gated behind `ui::color_enabled()` (TTY **and** `NO_COLOR` unset). When stdout is piped or `NO_COLOR` is set, the output falls back to the original plain list, **byte-for-byte unchanged**, and the JSON/CSV paths are untouched. Verified by diffing pre- vs post-change output:

- `sighthound sample | cat` — identical
- `NO_COLOR=1 sighthound sample` — identical
- `sighthound sample -o json` / `-o csv` — identical

## No new dependencies

Braille is packed by hand (`U+2800` base), in the spirit of the dependency-free `ui.rs`. `Cargo.toml` / `Cargo.lock` are untouched. Bars are padded to a fixed 10 glyphs so the count column stays aligned; deterministic ordering (BTreeMap + existing sorts) is preserved.

## Tests

Adds unit tests for the pure helpers in `output.rs`:
- `braille_bar` — the char-count == width alignment invariant (across widths/fracs), fill math, out-of-range clamping, and partial-cell behaviour.
- `severity_rank` — ordering including the unknown case.

Full suite green; no existing tests changed.

## Try it

```
cargo build --release
./target/release/sighthound <a directory with a few findings>
```

Any tree with findings across several files / severities will show both visuals in a colour terminal; piping it (`| cat`) shows the unchanged plain output.

Scope: a single file, `src/scanner/output.rs`.

### DIFF
--- a/src/scanner/output.rs
+++ b/src/scanner/output.rs
@@ -37,9 +37,9 @@ fn braille_bar(frac: f64, width: usize) -> String {
     let mut bar = String::with_capacity(width * 3);
     for cell in 0..width {
         let mut code = 0u32;
-        for col in 0..2 {
+        for (col, &mask) in FULL_COL.iter().enumerate() {
             if cell * 2 + col < lit {
-                code |= FULL_COL[col];
+                code |= mask;
             }
         }
         bar.push(char::from_u32(0x2800 + code).unwrap_or(' '));
@@ -115,9 +115,7 @@ pub fn print_summary(findings: &[Finding], duration: std::time::Duration) {
         *finding_types.entry(finding.finding_type.clone()).or_insert(0) += 1;
         *file_counts.entry(finding.file.clone()).or_insert(0) += 1;
         // Track the worst severity seen in each file, for the heat-bar colour.
-        let worst = file_worst
-            .entry(finding.file.clone())
-            .or_insert_with(|| sev.clone());
+        let worst = file_worst.entry(finding.file.clone()).or_insert_with(|| sev.clone());
         if severity_rank(&sev) > severity_rank(worst) {
             *worst = sev;
         }
@@ -142,10 +140,8 @@ pub fn print_summary(findings: &[Finding], duration: std::time::Duration) {
     // keeps just the line above, unchanged).
     if crate::ui::color_enabled() {
         const WIDTH: usize = 24;
-        let counts: Vec<usize> = severity_order
-            .iter()
-            .map(|s| severity_counts.get(*s).copied().unwrap_or(0))
-            .collect();
+        let counts: Vec<usize> =
+            severity_order.iter().map(|s| severity_counts.get(*s).copied().unwrap_or(0)).collect();
         if counts.iter().sum::<usize>() > 0 {
             let cells = severity_mix_cells(&counts, WIDTH);
             let mut mix = String::new();
--- a/src/scanner/output.rs
+++ b/src/scanner/output.rs
@@ -47,6 +47,53 @@ fn braille_bar(frac: f64, width: usize) -> String {
     bar
 }
 
+/// Allocate `width` bar cells across severities in proportion to `counts`,
+/// giving every severity with a non-zero count at least one cell while never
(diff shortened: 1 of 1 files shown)


asadeddin (commented):
@buttonmonkey 

Beautiful UI/UX contribution. 

A couple of quick questions:
1- Any concern over accessibility for color blindness? 
2- Most affected files, does it carry over the color scheme or will it always be red? 

Image

buttonmonkey:
1. Potentially, the bars are mapped to your classification colors: critical, high, medium, low. Not sure of how to mitigate that as I don't know how to create a reliable pattern fill or some equivalent that would work in this setting. If I were to find that pattern, it would require changing your color values to match the pattern fills.
2. The bars should equate to the classification types. Only the same color if they are all of the same type. (ex. all critical)

image

src/scanner/output.rs
```diff
@@ -49,6 +90,29 @@ pub fn print_summary(findings: &[Finding], duration: std::time::Duration) {
...
+        let total: usize = severity_counts.values().sum();
+        if total > 0 {
+            const WIDTH: usize = 24;
+            let mut mix = String::new();
+            for sev in severity_order.iter() {
+                if let Some(&count) = severity_counts.get(*sev) {
+                    if count == 0 {
+                        continue;
+                    }
+                    let cells = ((count * WIDTH) as f64 / total as f64).round().max(1.0) as usize;
```
juangaitanv: should we allocate the 24 cells cumulatively so per-segment rounding and `.max(1.0)` cannot exceed `WIDTH`? Counts of 997/1/1/1 render 27 cells.

buttonmonkey: Here's the overflow fix; I kept minorities visible so the bar matches the tally above it — happy to switch to plain cumulative if you'd prefer the bar strictly proportional.

PR #38, opened 2026-07-10
fix: add --fail-on-severity / --error-on-findings CI gate flags (#36)

### LINKED ISSUE #36: [Bug] Scanner always exits 0 even when vulnerabilities are found (no way to gate CI)
## Description

sighthound returns exit code 0 regardless of how many (or how severe) the findings are. There is no --fail-on-severity / threshold flag either. This makes it impossible to use as a CI/PR gate without a wrapper script that parses the JSON output and computes an exit code externally.

## Reproduction

Create `vuln.py`:
```py
import os

def run(cmd):
    os.system(cmd) 
```

Run:
```
sighthound --simple-analysis vuln.py python ; echo "exit code: $?"
```

Expected: for CI usage, a non-zero exit when findings meet a configurable threshold, like:
```
sighthound --fail-on-severity high .   # should exit 1
```

Actual: the scanner prints the "Critical" finding but prints `exit code: 0`. The same holds for `--output-format json`. In `src/main.rs`, `main()` unconditonally returns `Ok(())` after printing findings, so the number and severity of findings never affects the exit status.

I would suggest adding a flag like `--fail-on-severity <level>` (as above). Or `--error-on-findings`. This would map to a non-zero exit code when any finding at/above the given severity is present. This is the main blocker for adopting sighthound org-wide in a PR/CI pipeline. 

## Environment

- Sighthound version / commit (`sighthound --version` or git SHA): `sighthound 0.1.1` (built from `main`, `cargo build --release`)
- OS and architecture: macOS 26.5.2, `arm64`

Adityakk9031:
Closes #36

## Problem

`sighthound` always exits 0 regardless of findings severity, making it
impossible to use as a hard gate in CI without a wrapper that parses JSON
output externally.

## Solution

Two new flags control exit behaviour:

- `--fail-on-severity <SEVERITY>` — exits 1 if any finding is at or above
  the given level (`critical`, `high`, `medium`, `low`). Severity is
  compared ordinally so `--fail-on-severity high` also catches `critical`.
- `--error-on-findings` — exits 1 if any finding is reported, regardless
  of severity.

Invalid values (e.g. `--fail-on-severity bogus`) are rejected early with a
clear error message and a non-zero exit before scanning begins.

Example CI usage:

```yaml
- run: sighthound --fail-on-severity high .

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -184,22 +184,25 @@ fn main() -> Result<()> {
     output_findings(&cli, &findings, duration)?;
 
     // Check if we need to fail (exit 1) based on severity or findings count
-    let should_fail = if cli.error_on_findings && !findings.is_empty() {
-        true
+    let failure_reason = if cli.error_on_findings && !findings.is_empty() {
+        Some("findings found (--error-on-findings is set)".to_string())
     } else if let Some(ref fail_severity) = cli.fail_on_severity {
         let target_num = severity_to_num(fail_severity);
-        findings.iter().any(|f| severity_to_num(&f.severity) >= target_num)
+        let has_offending_finding = findings.iter().any(|f| severity_to_num(&f.severity) >= target_num);
+        if has_offending_finding {
+            Some(format!("findings at or above severity threshold ({})", fail_severity))
+        } else {
+            None
+        }
     } else {
-        false
+        None
     };
 
-    if should_fail {
-        let threshold = cli
-            .fail_on_severity
-            .as_deref()
-            .map(|s| format!(" (threshold: {})", s))
-            .unwrap_or_default();
-        sighthound::ui::error_line(&format!("exiting non-zero: findings above threshold{}", threshold));
+    if let Some(reason) = failure_reason {
+        sighthound::ui::error_line(&format!("exiting non-zero: {}", reason));
+        use std::io::{self, Write};
+        let _ = io::stdout().flush();
+        let _ = io::stderr().flush();
         std::process::exit(1);
     }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -188,7 +188,8 @@ fn main() -> Result<()> {
         Some("findings found (--error-on-findings is set)".to_string())
     } else if let Some(ref fail_severity) = cli.fail_on_severity {
         let target_num = severity_to_num(fail_severity);
-        let has_offending_finding = findings.iter().any(|f| severity_to_num(&f.severity) >= target_num);
+        let has_offending_finding =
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by asadeddin** [comment]
This is great and makes sense. Thanks for the PR! I'm tagging our reviewers on it.

File: src/main.rs [hunk 1]
```diff
@@ -172,9 +183,39 @@ fn main() -> Result<()> {
...
+    };
+
+    if should_fail {
+        let threshold = cli
+            .fail_on_severity
+            .as_deref()
+            .map(|s| format!(" (threshold: {})", s))
+            .unwrap_or_default();
+        sighthound::ui::error_line(&format!("exiting non-zero: findings above threshold{}", threshold));
+        std::process::exit(1);
```
juangaitanv: could we report which gate fired here? `--error-on-findings` can set `should_fail` even when no finding is above the severity threshold, so the current message is misleading.

tests/integration/integration_tests.rs
```diff
@@ -437,4 +437,64 @@ def malicious_functions():
...
+            .status()
+            .expect("failed to run sighthound");
+        assert_eq!(status.code(), Some(1));
+
+        // critical threshold — os.system is Critical, so this should trigger
+        let status = std::process::Command::new(&bin_path)
+            .args(&[file_path, "python", "--simple-analysis", "--fail-on-severity", "critical"])
+            .status()
+            .expect("failed to run sighthound");
+        assert_eq!(status.code(), Some(1));
```
juangaitanv: The embedded `os.system` rule is High, so it does not meet the `critical` threshold. This assertion should expect success, use `--fail-on-severity high`, or scan a fixture with a Critical finding.

src/main.rs
yhoztak: In Rust, exit bypasses destructors; any buffered data in `stdout`’s `LineWriter` (especially when piped/redirected in CI) is dropped. If output_findings emits JSON/SARIF, the output can be silently truncated, making it unparsable for CI consumers, so we want to explicitly flush before exiting:
```
use std::io::{self, Write};
// ...
io::stdout().flush().unwrap();
io::stderr().flush().unwrap();
std::process::exit(1);
```

Adityakk9031:
@juangaitanv please check it

juangaitanv (approved):
lgtm

Ibrahimrahhal (changes requested):
Linting check is failing

Adityakk9031:
@Ibrahimrahhal please check it

Ibrahimrahhal (approved):
Looks good, Thanks for contributing!

PR #39, opened 2026-07-10
fix: don't flag parameterized cursor.execute() as SQL injection

### LINKED ISSUE #35: [Bug] parameterized cursor.execute(... %s ...) flagged as "Critical SQL Injection"
## Description

The Python SQL-injection rule python-sql-raw-001 flags fully parameterized cursor.execute() calls as Critical SQL Injection. The pattern *cursor.execute(*%* matches the %s/%d bind-parameter placeholders used by DB-API drivers (psycopg2, mysqlclient, etc.), which are the safe, recommended way to pass values. This produces a high false-positive rate on real Django/psycopg codebases.

The same rule also flags psycopg's safe SQL-composition API (psycopg.sql.SQL(...).format(...)), which is designed to prevent injection.

## Reproduction

Create `repro.py`:
```py
import psycopg2

def get_rows(cursor, firm_id):
    # Correct, safe parameterization — value is bound by the driver, not interpolated.
    cursor.execute(
        "SELECT table_name, object_id FROM ccpa_objects "
        "WHERE firm_id = %s ORDER BY table_name",
        [firm_id],
    )
    return cursor.fetchall()
```

then:
```
sighthound --simple-analysis repro.py python
```

Expected: no finding, as it's a parametrized query (canonically safe, and widely used in our codebase). A true positive would require the value to be interpolated into the query string.

Actual: Reported as "Critical" via `python-sql-raw-001` because the pattern `*cursor.execute(*%*` matches the `%s` placeholder.

I found this in `rules/python/sql_injection.ron`:
```
"*cursor.execute(*%*",
```

I would suggest distinguishing interpolation from bind parameters. For example, when `%` is used as a string-format operator on the query, not when `%s/%(name)s` appears inside the literal. Or treat a call that passes a second argument to `execute()` as parametrized. You could also just exclude `psycopg.sql.SQL(...).format(...)` or  `psql.SQL(...).format(...)` from the `.format(` pattern. 

I measured ~60% of this rule's findings on a real, production Python codebase (~5k files). 

## Environment

- Sighthound version / commit (`sighthound --version` or git SHA): `sighthound 0.1.1` built from `main` + `cargo build --release`
- OS and architecture: macOS 26.5.2, arm64

mvanhorn:
## Summary

The `python-sql-raw-001` glob `*cursor.execute(*%*` treated any `%` in the call text as string interpolation, so canonical DB-API parameter binding (`cursor.execute("... WHERE firm_id = %s", [firm_id])`) was reported as Critical SQL Injection. The reporter measured ~60% of this rule's findings on a ~5k-file production codebase as this false positive (#35).

This replaces the glob with a `regex:` pattern (Pattern Type 4 in the rule guide) that only fires when `%` is a formatting operator after a complete query expression, a closing quote or a bare identifier, instead of matching `%s`/`%(name)s` placeholders inside the string literal. Placeholders bound via `execute()`'s second argument now pass clean; every interpolation form still flags.

Regression coverage both ways, wired into the acceptance harness:
- `false_positive_floor.feature`: parameterized `%s`, `%(name)s`, and a bound `f"%{name}%"` LIKE argument produce zero findings
- `python_injection.feature`: spaced `"..." % x`, unspaced `"..."%x`, variable-held `q % x`, f-string, concatenation, and `.format()` all still report SQL Injection

One part of #35 is intentionally not covered here: the rule's `*cursor.execute(*.format(*` pattern still flags psycopg's safe `sql.SQL(...).format(...)` composition API. Search rules have no pattern-level exclude mechanism, so that case needs its own treatment rather than being bundled into this fix.

Verified with a rebuilt binary against the issue's exact repro: `Found 0 vulnerabilities`. `cargo test --test acceptance` passes 15/15 scenarios.

## Related issue

Closes #35

## Checklist

- [x] `cargo build --release` succeeds
- [x] Ran any test harness relevant to this change (note results; suite under repair) - `cargo test --test acceptance`: 15 scenarios / 52 steps pass, including the 7 new scenarios covering this rule
- [x] Updated docs/rules where relevant - the change is to `rules/python/sql_injection.ron` itself, with an inline comment explaining what the pattern matches

### DIFF
--- a/rules/python/sql_injection.ron
+++ b/rules/python/sql_injection.ron
@@ -10,7 +10,8 @@
             mode: "search",
             patterns: Some([
                 "execute_raw_query(",
-                "*cursor.execute(*%*",
+                // Match the % operator after a complete query expression, not DB-API placeholders.
+                "regex:(?s)cursor\\.execute\\(\\s*\\(*\\s*(?:(?:[rRuUbBfF]{0,2})?(?:\"\"\".*?\"\"\"|'''.*?'''|\"(?:\\\\.|[^\"\\\\])*\"|'(?:\\\\.|[^'\\\\])*')|[A-Za-z_][A-Za-z0-9_.]*)\\s*\\)*\\s*%\\s*[^\\s,)]",
                 "*cursor.execute(*.format(*",
                 "*cursor.execute(f\"*",
                 "*cursor.execute(f'*",
--- a/tests/test_files/python/sql_percent_unspaced.py
+++ b/tests/test_files/python/sql_percent_unspaced.py
@@ -1,0 +1,2 @@
+def load_user(cursor, user_id):
+    cursor.execute("SELECT * FROM users WHERE id = %s"%user_id)
--- a/tests/features/false_positive_floor.feature
+++ b/tests/features/false_positive_floor.feature
@@ -7,3 +7,8 @@ Feature: False-positive floor
     Given a staged benign Python module "calculator.py"
     When I scan the staging directory as "python" with the production rules
     Then no findings should be reported
+
+  Scenario: Parameterized cursor execution produces zero findings
+    Given a staged copy of the fixture "tests/test_files/python/sql_parameterized_execute.py" as "parameterized_query.py"
+    When I scan the staging directory as "python" with the production rules
+    Then no findings should be reported
--- a/tests/features/python_injection.feature
+++ b/tests/features/python_injection.feature
@@ -8,3 +8,17 @@ Feature: Python injection detection
     When I scan the staging directory as "python" with the production rules
     Then the findings should include a "Command Injection" finding in "mixed_case.py"
     And the findings should include a "SQL Injection" finding in "mixed_case.py"
+
+  Scenario Outline: Dynamically formatted cursor execution is detected
+    Given a staged copy of the fixture "<fixture>" as "unsafe_query.py"
+    When I scan the staging directory as "python" with the production rules
+    Then the findings should include a "SQL Injection" finding in "unsafe_query.py"
+
+    Examples:
+      | fixture                                                    |
+      | tests/test_files/python/sql_percent_spaced.py              |
+      | tests/test_files/python/sql_percent_unspaced.py            |
+      | tests/test_files/python/sql_percent_variable.py            |
+      | tests/test_files/python/sql_fstring_interpolation.py        |
+      | tests/test_files/python/sql_concatenation.py                |
+      | tests/test_files/python/sql_format_interpolation.py         |
--- a/tests/test_files/python/sql_concatenation.py
+++ b/tests/test_files/python/sql_concatenation.py
@@ -1,0 +1,2 @@
+def load_user(cursor, user_id):
+    cursor.execute("SELECT * FROM users WHERE id = " + user_id)
--- a/tests/test_files/python/sql_format_interpolation.py
+++ b/tests/test_files/python/sql_format_interpolation.py
@@ -1,0 +1,2 @@
+def load_user(cursor, user_id):
+    cursor.execute("SELECT * FROM users WHERE id = {}".format(user_id))
--- a/tests/test_files/python/sql_fstring_interpolation.py
+++ b/tests/test_files/python/sql_fstring_interpolation.py
@@ -1,0 +1,2 @@
+def load_user(cursor, user_id):
+    cursor.execute(f"SELECT * FROM users WHERE id = {user_id}")
--- a/tests/test_files/python/sql_parameterized_execute.py
+++ b/tests/test_files/python/sql_parameterized_execute.py
@@ -1,0 +1,13 @@
+def load_firms(cursor, firm_id, firm_name):
+    cursor.execute(
+        "SELECT * FROM firms WHERE firm_id = %s",
+        [firm_id],
+    )
+    cursor.execute(
+        "SELECT * FROM firms WHERE name = %(name)s",
+        {"name": firm_name},
+    )
+    cursor.execute(
+        "SELECT * FROM firms WHERE name LIKE %s",
+        [f"%{firm_name}%"],
+    )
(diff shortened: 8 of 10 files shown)


asadeddin:
Thank you for the second PR. The rule change makes sense. I've tagged some of our reviews to chime in.

tests/test_files/python/sql_percent_unspaced.py
```diff
@@ -0,0 +1,2 @@
+def load_user(cursor, user_id):
```
yhoztak, line 1: Can we add these to the same sql test file for readability and coverages? 

like: 

```
def percent_spaced(cursor, user_id):
    cursor.execute("SELECT * FROM users WHERE id = %s" % user_id)


def percent_unspaced(cursor, user_id):
    cursor.execute("SELECT * FROM users WHERE id = %s"%user_id)


def percent_variable(cursor, query, user_id):
    cursor.execute(query % user_id)


def fstring_interpolation(cursor, user_id):
    cursor.execute(f"SELECT * FROM users WHERE id = {user_id}")


def string_concatenation(cursor, user_id):
    cursor.execute("SELECT * FROM users WHERE id = " + user_id)


def format_interpolation(cursor, user_id):
    cursor.execute("SELECT * FROM users WHERE id = {}".format(user_id))

```

rules/python/sql_injection.ron
```diff
@@ -10,7 +10,8 @@
             mode: "search",
             patterns: Some([
                 "execute_raw_query(",
-                "*cursor.execute(*%*",
+                // Match the % operator after a complete query expression, not DB-API placeholders.
+                "regex:(?s)cursor\\.execute\\(\\s*\\(*\\s*(?:(?:[rRuUbBfF]{0,2})?(?:\"\"\".*?\"\"\"|'''.*?'''|\"(?:\\\\.|[^\"\\\\])*\"|'(?:\\\\.|[^'\\\\])*')|[A-Za-z_][A-Za-z0-9_.]*)\\s*\\)*\\s*%\\s*[^\\s,)]",
```
yhoztak, line 14: We try to avoid very complex regex like this as it may become too specific and hard to understand it. 
We could have separate AST rule here instead. 

As an example: 
```
(
    rules: [
        // Covers `"..." % x` and `"..." + x` passed to cursor.execute.
        // Parameterized execute("...%s", params) has a string as arg0, not
        // a binary_operator, so it is not matched.
        (
            id: Some("python-sql-raw-binop"),
            name: Some("Raw SQL Execution (binary-built query)"),
            category: Some("database"),
            mode: "search",
            patterns: Some([
                "cursor.execute",
            ]),
            conditions: Some([
                (
                    field: "argument",
                    operator: "matches",
                    value: "",
                    condition_type: Some("has_argument"),
                    argument_position: Some(0),
                    node_type: Some("binary_operator"),
                ),
            ]),
            finding_type: Some("SQL Injection"),
            severity: Some("Critical"),
            confidence: Some("Medium"),
            cwe_id: Some("cwe-89"),
            description: Some("Raw SQL query executed from a dynamically-built string (string formatting/concatenation)"),
            file_types: Some((
                extensions: Some([".py"])
            )),
            tags: Some(["sql", "injection", "database"])
        ),

        // Remaining inline string-building shapes the binary-operator rule
        // does not cover (f-strings, str.format, raw-query helpers).
        (
            id: Some("python-sql-raw-001"),
            name: Some("Raw SQL Execution"),
            category: Some("database"),
            mode: "search",
            patterns: Some([
                "execute_raw_query(",
                "*cursor.execute(*.format(*",
                "*cursor.execute(f\"*",
                "*cursor.execute(f'*",
                "*.executescript(*",
            ]),
            finding_type: Some("SQL Injection"),
            severity: Some("Critical"),
            confidence: Some("Medium"),
            cwe_id: Some("cwe-89"),
            description: Some("Raw SQL query executed from a dynamically-built string (string formatting/concatenation)"),
            file_types: Some((
                extensions: Some([".py"])
            )),
            tags: Some(["sql", "injection", "database"])
        ),

        // Taint: user input flowing into a SQL execute sink (catches variable-built
        // queries that aren't visible as a literal at the call site).
        (
            id: Some("python-sql-taint-001"),
            name: Some("SQL Injection Taint Flow"),
            category: Some("database"),
            mode: "taint",
            sources: Some([
                "request.",
                "input(",
                "raw_input(",
                "flask.request",
            ]),
            sinks: Some([
                "execute(",
                "executemany(",
                "executescript(",
            ]),
            sanitizers: Some([
                "parameterized",
            ]),
            finding_type: Some("SQL Injection"),
            severity: Some("Critical"),
            confidence: Some("High"),
            cwe_id: Some("cwe-89"),
            description: Some("User input flows to SQL execution without parameterization"),
            file_types: Some((
                extensions: Some([".py"])
            )),
            tags: Some(["sql", "injection", "taint", "database"])
        ),
    ]
)

```

juangaitanv, line 14: The trailing `%\s*[^\s,)]` excludes `(`, so tuple formatting such as `cursor.execute("SELECT ... %s" % (user_id,))` no longer matches. This leaves a common SQL-injection pattern undetected; allow `(` and add a tuple-RHS fixture.

tests/test_files/python/sql_percent_unspaced.py
yhoztak: We are cleaning up test folder structures starting from python and PR is up here fyi https://github.com/Corgea/Sighthound/pull/43
My apology for not having a clean structure to follow

yhoztak: https://github.com/Corgea/Sighthound/pull/43 is now merged. If there's a new fixture files for test cases, can you move them under https://github.com/Corgea/Sighthound/tree/main/tests/test_files/python/fixtures ?
As an example for SQL Injection, add please add new functions to https://github.com/Corgea/Sighthound/blob/main/tests/test_files/python/fixtures/sql_injection_variants.py

Issue #40, opened 2026-07-11
[Architecture Improvement] More strict scoping of attacker-controlled input


freakynit:
## Description

A lot of false positives. Like > 90%

Can't share the results here since it was run on a private repo, but, can share privately (with relevant redactions).

A few I can list:

1. False positive: keys come from a fixed allowlist, not user-controlled property names.
2. False positive: keys are internal entity types; values are bounded comparison data.
3.  False positive: property names are fixed and values must be booleans.
4. False positive: only five fixed keys are copied.
...
...

tinovyatkin:
@freakynit @mvanhorn this FP class is interesting — the examples (@freakynit's "keys come from a fixed allowlist," "values must be booleans") aren't really syntax problems, they're *provenance* problems: the analyzer can't tell a value that's pinned to a constant/allowlist from one that's actually attacker-controlled. And PR #39 (swapping the `cursor.execute` glob for a regex in #35) is the same story one level down — you're refining a textual pattern where what you'd want is "is this argument tainted?"

That distinction is hard to make on a tree-sitter CST because it carries no scope, types, or expression provenance — so taint ends up approximated with string/AST-shape patterns, and constants look the same as tainted values. Getting under that FP floor eventually needs a real dataflow model, and that model needs a parse tree with precedence, scoping, and typed expressions as its substrate.

Worth knowing for the languages where it's tractable: for **Java and C#** specifically, ANTLR's `grammars-v4` has mature semantic grammars, and I maintain [antlr-rust-runtime](https://github.com/ophi-dev/antlr-rust-runtime) — a clean-room ANTLR v4 runtime for Rust (stable, BSD-3, passes the upstream conformance suite). Not suggesting you rip out tree-sitter across all 8 languages — it's fine for a lot of what you do — but for the JVM/C# rules where these FPs bite, a real parse tree is the foundation a proper source→sink taint pass can build on instead of pattern-tuning. Your `tests/strictness` taint fixtures + `false_positive_floor` suite look like they'd make experimenting with a deeper backend on one language low-risk. Happy to help if you ever want to try it.

freakynit:
@tinovyatkin Thanks for the detailed explanation. I'm not familiar with compiler, parser, or static-analysis internals, so I needed some help (chatgpt helped) unpacking it, but the distinction is much clearer now.

[...]

PR #41, opened 2026-07-11
fix: resolve SQL injection and taint analysis false positives (#35, #40)

Adityakk9031:
## Description
This PR resolves major false-positive issues in the scanner's taint-tracking engine and Python rule matching engines. It combines fixes for SQL injection dynamic string formatting and tightens dataflow tracing across PHP and Python.

**Fixes:**  #40

### Key Changes
1. **Sanitizer Wrapper Detection**: Added `is_sanitizer_wrapper` to `TaintRuleDeduplicator` and updated `classify_assignment_rhs`. This allows the dataflow tracer to correctly recognize and short-circuit when an entire expression is wrapped in a known sanitizer (e.g., `escapeshellarg(...)` in PHP), preventing inner variables from triggering false positives.
2. **Structural Node Scoping**: Refactored `scan_file_with_taint_rules` to apply container-skipping node filters (like ignoring `function_definition`) *only* to the variable tracing loop. This ensures inline source-in-sink pattern matching works robustly while avoiding misattributing taint flows to entire function bodies.
3. **PHP Variable Parsing**: Fixed an issue in `find_assignment_source` where stripping the `$` sigil caused the scanner to miss assignments. The tracer now properly resolves both `$variable_name` and `variable_name` patterns.
4. **Python DB-API Parameterized Queries**: Handled parameter placeholders (like `%s` and `%d`) that previously caused dynamic string formatting rules to incorrectly flag SQL injections on safe `cursor.execute()` calls.
5. **Collection & Loop Tracing**: Implemented classification for safe collection literals (e.g., `[...]` / `{...}`) and dictionary access expressions (`VAR.get(key)` / `VAR[key]`), resolving them to `Unknown` instead of defaulting to `Tainted`. It also ensures `for` loops map back to the actual iterable instead of `FunctionParameter`.

### Testing
- Validated all 24/24 strictness regression tests (including `allowlist_bounded_access_produces_no_taint_findings`).
- Ran `make check` and `make ci` locally to ensure no architectural/coverage drifts or complexity issues.

### DIFF
--- a/src/scanner/dataflow.rs
+++ b/src/scanner/dataflow.rs
@@ -260,6 +260,16 @@ impl DataFlowTracer {
         file_line: usize,
         rule_deduplicator: &TaintRuleDeduplicator,
     ) -> VariableSource {
+        // If the entire RHS is wrapped in a known sanitizer call (e.g.
+        // `escapeshellarg($_GET["cmd"])`), it is safe regardless of what it contains.
+        if rule_deduplicator.is_sanitizer_wrapper(rhs) {
+            log::debug!("[COMPUTE_VARIABLE_SOURCE] Sanitizer-wrapped assignment: '{}'", rhs);
+            return VariableSource::KnownSafe {
+                reason: "sanitizer wrapper".to_string(),
+                line: file_line,
+            };
+        }
+
         match self.classify_value_source(rhs, rule_deduplicator) {
             ValueSourceClassification::Safe(reason) => {
                 log::debug!("[COMPUTE_VARIABLE_SOURCE] Proven-safe assignment: {}", reason);
@@ -311,7 +321,23 @@ impl DataFlowTracer {
             let line = line.trim();
             // Note: augmented assignments (`x += ...`, `x -= ...`) are intentionally
             // not matched by this `{} =` guard and are not handled by this path.
-            if !line.starts_with(&format!("{} =", variable_name)) {
+            //
+            // Also match PHP-style `$varname = ...` assignments: `extract_php_variables`
+            // strips the leading `$` sigil, so the extracted name is `"cmd"` while the
+            // actual source line reads `$cmd = $_GET["cmd"]`.
+            //
+            // Also match JS/TS `var x = ...`, `let x = ...`, `const x = ...` declarations.
+            let plain_prefix = format!("{} =", variable_name);
+            let php_prefix = format!("${} =", variable_name);
+            let js_var_prefix = format!("var {} =", variable_name);
+            let js_let_prefix = format!("let {} =", variable_name);
+            let js_const_prefix = format!("const {} =", variable_name);
+            if !line.starts_with(&plain_prefix)
+                && !line.starts_with(&php_prefix)
+                && !line.starts_with(&js_var_prefix)
+                && !line.starts_with(&js_let_prefix)
+                && !line.starts_with(&js_const_prefix)
+            {
                 continue;
             }
             // Split on the FIRST '=' only so the full RHS is preserved even when
@@ -347,7 +373,7 @@ impl DataFlowTracer {
     }
 
     fn compute_variable_source(
-        &self,
+        &mut self,
         file_path: &str,
         function_name: &str,
         variable_name: &str,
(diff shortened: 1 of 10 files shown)


juangaitanv (commented):
thank you for your contribution 🙌!

a couple of PR-level questions:
- could we split the #35 sql rule change into its own PR so it can merge independently of the taint-engine change?
- is the two-phase to backward-tracer swap required for the #40 allowlist fixes, or could those land as a post-filter on the existing tracker?

src/scanner/scanning_logic.rs
```diff
@@ -1185,39 +1186,172 @@ impl ScanningLogic {
             return findings;
         }
 
-        // Create rule deduplicator to prevent cartesian product problems
+        // Create rule deduplicator
         let rule_deduplicator = TaintRuleDeduplicator::new(&applicable_rules);
+        let mut data_flow_tracer = DataFlowTracer::new();
```
juangaitanv, line 807: should we drop the two-phase functions and the now-unused `flow_tracker` members now that nothing calls them, since `make ci` runs clippy with `-D warnings`?

src/scanner/scanning_logic.rs
```diff
@@ -1185,39 +1186,172 @@ impl ScanningLogic {
...
-            Self::collect_taint_sink_findings_for_node(
-                node,
-                &ctx,
-                &mut flow_tracker,
-                &mut findings,
-            );
+                // Skip structural container nodes for indirect variable-based taint tracing.
+                // Their text spans the entire function body, so tracing any variable extracted
+                // from them attributes the finding to the function's START line instead of the
+                // actual sink expression line. The inline check above still runs for these nodes.
```
juangaitanv: should the inline source-in-sink check also skip `function_definition` nodes, e.g. gate it on `is_actionable_sink_node` like the old path?

src/scanner/taint_utils.rs
```diff
@@ -40,11 +43,47 @@ impl TaintRuleDeduplicator {
...
 
+    /// Returns `true` when the outermost function call in `expr` is a known sanitizer.
+    ///
+    /// Matches either:
+    /// * a sanitizer pattern that is itself a call prefix (e.g. `"escapeshellarg("`), or
+    /// * any sanitizer name that ends with `(` and whose base name matches the leading
+    ///   identifier in `expr`.
+    ///
+    /// Example: `is_sanitizer_wrapper("escapeshellarg($_GET[\"cmd\"])")` → `true`
+    pub(crate) fn is_sanitizer_wrapper(&self, expr: &str) -> bool {
```
juangaitanv, line 73: should `is_sanitizer_wrapper` match only the current sink rule's sanitizers rather than the pool across all applicable rules?

src/scanner/dataflow.rs
```diff
@@ -489,7 +560,36 @@ impl DataFlowTracer {
         ValueSourceClassification::Unknown
     }
 
-    fn is_safe_literal_expression(&self, expression: &str) -> bool {
+    /// Returns true when `expr` looks like a dict/collection access: `VAR.get(...)`,
+    /// `VAR[...]`, or `VAR.get(...)` with a default. These shapes are safe when VAR is
+    /// a locally defined constant collection.
+    fn is_dict_access_expression(expression: &str) -> bool {
```
juangaitanv, line 586: for `VAR[...]`, could a genuinely tainted subscript resolve to `Unknown`?

tests/strictness/false_positive_regressions.rs
```diff
@@ -23,3 +23,49 @@ fn cleartext_transmission_regression_fixture() {
...
+        "parameterized queries should not be flagged, got {} finding(s): {:?}",
+        sql_findings.len(),
+        sql_findings.iter().map(|f| (f.line, f.snippet.as_str())).collect::<Vec<_>>()
+    );
+}
+
+/// Regression for #40: allowlist-bounded key access and constant dict lookups must not
+/// produce taint findings even when a real taint source (request.json) is present.
+#[test]
+fn allowlist_bounded_access_produces_no_taint_findings() {
```
juangaitanv, line 52: should we add a true-positive case asserting a real cross-statement flow is still reported at the sink line?

PR #44, opened 2026-07-14
fix: use AST-based rules to avoid SQL injection false positives on parameterized queries (#35)

### LINKED ISSUE #35: [Bug] parameterized cursor.execute(... %s ...) flagged as "Critical SQL Injection"

Adityakk9031:
## Description
Closes #35

## Problem
The Python SQL injection rule (`python-sql-raw-001`) caused high false-positive rates on legitimate DB-API 2.0 parameterized queries. The broad glob pattern `cursor.execute(%*` incorrectly flagged safe placeholders like `%s` / `%d`, e.g.:

```python
cursor.execute("SELECT * FROM users WHERE id = %s", [user_id])  # safe — was flagged
```

### 1. AST-based binary-operator rule (`python-sql-raw-binop`)
A dedicated rule that checks whether the 0th argument to `cursor.execute` is a `binary_operator` AST node. This natively catches all unsafe `"..." % x` and `"..." + x` interpolation without being triggered by bind-parameter tuples/lists passed as a second argument.

### 2. First-argument–anchored f-string / `.format` rule (`python-sql-raw-format`)
A new rule that matches f-strings and `.format` calls only when they appear as the query argument (position `0`) of `cursor.execute`. This prevents parameterized calls like `cursor.execute("...%s", [f"user_{uid}"])` from being flagged.

### 3. Narrowed sanitizer exclusions
The overly broad `*SQL*` sanitizer pattern was replaced with precise `*sql.SQL(*` and `*psql.SQL(*` patterns. This restores detection of real injections that happen to use SQL-named identifiers (e.g. `cursor.execute(SQL_QUERY % user_id)`).

### 4. Keyword-argument handling (`unwrap_keyword_argument`)
Centralized keyword-argument unwrapping into a single `unwrap_keyword_argument` helper so that the keyword form `cursor.execute(query="..." % uid)` is correctly flagged alongside the positional form.

### 5. Full-text context scoping
The `rule_needs_full_context` broadening was scoped exclusively to the database rule category, avoiding any engine-wide matching-semantic change to the other 43 rule files.

### 6. Cross-platform harness fix
`write_if_missing` in `harness.rs` was fixed with `#[cfg_attr(not(unix), allow(unused_variables))]` so the build compiles cleanly on both Unix and Windows (`E0425` is resolved).

## Verification

- `cargo test`: 105 tests pass (all platforms)
- 11 false positives removed vs. `main` on the Python fixture corpus
- 1 new true positive recovered (keyword f-string form)
- No lost detections
- Safe parameterized queries (`%s` bind params, keyword `query=` form, `psycopg sql.SQL(...)`) all scan clean
- Previously suppressed injections (`SQL_QUERY % user_id`, `MYSQL_AUDIT` concat, f-string with SQL-named variable) are now correctly flagged


Adityakk9031:
@juangaitanv check this

Adityakk9031:
@Ibrahimrahhal please check this

juangaitanv:
thank you again 🙏! i'll take a look asap. in the meantime, could you fix the CI / harness issues please?

rules/python/sql_injection.ron
```diff
@@ -1,20 +1,49 @@
...
+        // a binary_operator, so it is not matched.
+        (
+            id: Some("python-sql-raw-binop"),
+            name: Some("Raw SQL Execution (binary-built query)"),
+            category: Some("database"),
+            mode: "search",
+            patterns: Some([
+                "cursor.execute",
+            ]),
+            conditions: Some([
```
juangaitanv, line 14: should these conditions also hold on the `--language-filter`/`--code-type` path? `check_rule_against_node_with_taint_context` skips `check_ast_conditions`, and there every `cursor.execute` call matches, parameterized included.

rules/python/sql_injection.ron
```diff
@@ -1,20 +1,49 @@
...
+                (
+                    field: "argument",
+                    operator: "matches",
+                    value: "",
+                    condition_type: Some("has_argument"),
+                    argument_position: Some(0),
```
juangaitanv, line 20: does the position-0 check need to cover the keyword form `cursor.execute(sql="..." % x)` too?

rules/python/sql_injection.ron
```diff
@@ -1,20 +1,49 @@
...
+        // does not cover (f-strings, str.format, raw-query helpers).
         (
             id: Some("python-sql-raw-001"),
             name: Some("Raw SQL Execution"),
             category: Some("database"),
             mode: "search",
             patterns: Some([
                 "execute_raw_query(",
-                "*cursor.execute(*%*",
                 "*cursor.execute(*.format(*",
```
juangaitanv: could the `.format(` glob exclude psycopg's safe `sql.SQL(...).format(...)` composition mentioned in #35?

rules/python/sql_injection.ron
```diff
@@ -1,20 +1,49 @@
...
                 "*cursor.execute(f\"*",
```
juangaitanv: should the f-string and `.format` patterns still match with the `%`/`+` globs removed? `rule_needs_full_context` keyed on those for full-call-text matching, and `sql_formatting.py` lines 14 and 22 now scan clean on this branch.

tests/strictness/language_coverage.rs
```diff
@@ -223,7 +223,7 @@ fn include_test_fixtures_flag_enables_scanning_tests_directory() {
         with_flag.iter().any(|f| {
             f["file"]
                 .as_str()
-                .is_some_and(|path| path.contains("strictness_languages/go/unsafe.go"))
+                .is_some_and(|path| path.contains("unsafe.go"))
```
juangaitanv: do we still need the loosened path match in the sql-rule split, given it changes no scanner code and only one `unsafe.go` exists?

tests/test_files/python/sql_formatting.py
```diff
@@ -0,0 +1,22 @@
+def percent_spaced(cursor, user_id):
```
juangaitanv, line 1: should a strictness test assert these lines are flagged, e.g. via `scan_python_simple_with_rules` like `search_conditions.rs`?

tests/test_files/python/sql_formatting.py
```diff
@@ -0,0 +1,22 @@
...
+    cursor.execute("SELECT * FROM users WHERE id = %s" % user_id)
```
juangaitanv, line 2: could we also bring `sql_safe_parameterized.py` and the commented-out `dbapi_parameterized_queries_are_not_sql_injection` test over from #41 so the parameterized cases have a live assertion?

juangaitanv (commented):
thanks for following through on the split from #41 🙌, this is exactly the shape we hoped for: a self-contained rule change that can merge independently of the taint work. the AST `binary_operator` condition is the right way to distinguish interpolation from bind parameters, and it fixes the reported false positive in the default scan path.

one ask before give the go: could we back the verification claims with committed tests? nothing currently references `sql_formatting.py` or the new rule id, and scanning the fixture on this branch shows the f-string and `.format()` cases (lines 14 and 22) no longer produce findings. a strictness test would have caught that before review. #41 already has most of the pieces (`sql_safe_parameterized.py` and the commented-out `dbapi_parameterized_queries_are_not_sql_injection` test), so it's mostly a matter of carrying them over and enabling them.

details in the inline comments. happy to help iterate on this one.

Adityakk9031:
@juangaitanv please check

rules/python/sql_injection.ron
juangaitanv: addressed in ef1adf1, conditions are now checked in the enhanced-search path too. resolving.

rules/python/sql_injection.ron
juangaitanv: addressed via the keyword_argument unwrap in conditions.rs. resolving, one small follow-up on argument order in the new comments.

rules/python/sql_injection.ron
juangaitanv: addressed via the argument_not_sanitized exclusion. resolving, follow-up on the `*SQL*` pattern breadth in the new comments.

tests/test_files/python/sql_formatting.py
juangaitanv: addressed by raw_sql_injection_formatting_variants_are_flagged. resolving.

tests/test_files/python/sql_formatting.py
juangaitanv: addressed, fixture and test both carried over and enabled. resolving.

harness.rs
```diff
@@ -1119,7 +1119,7 @@ fi
 /// Write `path` with `content` only if it does not already exist. Returns true
 /// when it created the file. Never clobbers an existing file — preserves any
 /// local customization; `check_stop_hook_present` then reports the result.
-fn write_if_missing(path: &Path, content: &str, executable: bool) -> bool {
+fn write_if_missing(path: &Path, content: &str, _executable: bool) -> bool {
```
juangaitanv: shouldn't line 1133 pick up the rename too? on unix the `#[cfg(unix)]` block still reads `if executable` and `cargo build` fails with E0425, maybe `#[cfg_attr(not(unix), allow(unused_variables))]` instead.

rules/python/sql_injection.ron
```diff
@@ -1,22 +1,75 @@
...
+                ),
+                (
+                    field: "argument",
+                    operator: "not_contains",
+                    value: "",
+                    condition_type: Some("argument_not_sanitized"),
+                    patterns: Some([
+                        "*sql.SQL*",
+                        "*psql.SQL*",
+                        "*SQL*",
```
juangaitanv: should `*SQL*` narrow to maybe `*sql.SQL(*`/`*psql.SQL(*`? any argument containing SQL suppresses the finding: `cursor.execute(SQL_QUERY % user_id)` scans clean on this branch but is flagged on main.

rules/python/sql_injection.ron
```diff
@@ -1,22 +1,75 @@
...
             category: Some("database"),
             mode: "search",
             patterns: Some([
                 "execute_raw_query(",
-                "*cursor.execute(*%*",
                 "*cursor.execute(*.format(*",
-                "*cursor.execute(f\"*",
-                "*cursor.execute(f'*",
-                "*cursor.execute(*+*",
+                "*cursor.execute(*f\"*",
```
juangaitanv: does the extra `*` before `f"` need to match the params side too? `cursor.execute("... %s", [f"user_{uid}"])` is flagged on this branch, maybe `*cursor.execute(*=f"*` for the keyword form.

rules/python/sql_injection.ron
```diff
@@ -1,22 +1,75 @@
...
+            conditions: Some([
+                (
+                    field: "argument",
+                    operator: "not_contains",
+                    value: "",
+                    condition_type: Some("argument_not_sanitized"),
+                    patterns: Some([
+                        "*sql.SQL*",
+                        "*psql.SQL*",
+                        "*SQL*",
```
juangaitanv: does `*SQL*` need the same narrowing as line 31?

src/scanner/conditions.rs
```diff
@@ -66,10 +66,15 @@ pub fn check_has_argument_condition(
 
 /// Check if an argument node matches the given condition
 pub fn check_argument_matches(
-    arg: tree_sitter::Node,
+    mut arg: tree_sitter::Node,
     source: &[u8],
     condition: &Condition,
 ) -> bool {
+    if arg.kind() == "keyword_argument" {
```
juangaitanv: could the four `keyword_argument` unwraps share a helper, something like `unwrap_keyword_argument(arg)`?

src/scanner/conditions.rs
```diff
@@ -66,10 +66,15 @@ pub fn check_has_argument_condition(
...
+        if let Some(val_node) = arg.child_by_field_name("value") {
```
juangaitanv: does the position-0 lookup handle reversed keyword order, e.g. `cursor.execute(vars=(u, p), query="..." % x)`?

src/scanner/scanning_logic.rs
```diff
@@ -105,8 +105,14 @@ impl ScanningLogic {
             "=",
         ];
 
-        let check_pattern =
-            |pattern: &str| CONTEXT_INDICATORS.iter().any(|indicator| pattern.contains(indicator));
+        let check_pattern = |pattern: &str| {
+            CONTEXT_INDICATORS.iter().any(|indicator| pattern.contains(indicator))
+                || pattern.contains('.')
```
juangaitanv: could the full-text broadening land as its own PR, maybe as an explicit `match_full_text` rule flag? it flips 505 patterns across 43 rule files, well beyond the sql fix.

src/scanner/scanning_logic.rs
```diff
@@ -105,8 +105,14 @@ impl ScanningLogic {
...
+                || pattern.contains('*')
+                || pattern.contains('(')
```
juangaitanv: should a literal `require('database')` flag as dynamic code loading? it does on this branch via a revived pattern, so the broadened predicate may want a not-literal guard.

src/scanner/scanning_logic.rs
```diff
@@ -105,8 +105,14 @@ impl ScanningLogic {
...
+                || pattern.contains("f\"")
+                || pattern.contains("f'")
```
juangaitanv: do the rule-file contracts need updating with this, e.g. `command_injection.ron:4` still says bare dotted names match the function name?

tests/test_files/python/sql_formatting.py
```diff
@@ -0,0 +1,30 @@
...
+def format_interpolation(cursor, user_id):
+    cursor.execute("SELECT * FROM users WHERE id = {}".format(user_id))
+
+
+def keyword_percent_spaced(cursor, user_id):
+    cursor.execute(query="SELECT * FROM users WHERE id = %s" % user_id)
+
+
+def keyword_fstring_interpolation(cursor, user_id):
+    cursor.execute(query=f"SELECT * FROM users WHERE id = {user_id}")
```
juangaitanv, line 30: could we add cases whose names contain SQL, e.g. `cursor.execute(SQL_QUERY % user_id)`, so the sanitizer exclusion has regression cover?

juangaitanv (commented):
thanks for the fast, thorough turnaround, every thread from the last review came back addressed in code, and the new strictness tests are exactly the shape we asked for 🙌

I built this branch on macOS and ran the full suite plus comparison scans against main, three things worth a look before merge:

- `harness.rs` doesn't compile on unix: line 1133 still reads `if executable` after the parameter rename, so `cargo build` and `cargo test` fail with E0425 outside windows.
- the `*SQL*` sanitizer exclusion suppresses real injections: `cursor.execute(SQL_QUERY % user_id)` and a `"...MYSQL_AUDIT..." + user_id` concat scan clean on this branch but are flagged on main. narrowing to the psycopg patterns keeps all the new tests green.
- the `rule_needs_full_context` broadening flips 505 patterns across 43 rule files to full-call-text matching. on our fixture corpus the delta is small (+6/−1 findings, mostly revived true positives), which is encouraging, but it changes matching semantics engine-wide and would be easier to land as its own PR, with the rule-file comments that still describe function-name matching updated alongside.

details in the inline comments. I resolved the earlier threads since each one was addressed; the remaining questions live in the new comments. could we also update the PR description to cover the engine changes? they're now the biggest part of the diff.

Adityakk9031:
@juangaitanv have a look

Adityakk9031:
@juangaitanv any feedback

harness.rs
juangaitanv: still failing at 060f416: line 1133 reads `if executable` inside `#[cfg(unix)]` while the parameter is `_executable`, so unix builds hit E0425.

rules/python/sql_injection.ron
juangaitanv: still open at 060f416: `*SQL*` remains in the sanitizer list and `cursor.execute(SQL_QUERY % user_id)` still scans clean here while main flags it.

rules/python/sql_injection.ron
juangaitanv: still open at 060f416: the glob matches the whole call text, so the safe parameterized call with an f-string param is still flagged.

rules/python/sql_injection.ron
juangaitanv: still open at 060f416, same fix as the line-31 thread.

tests/test_files/python/sql_formatting.py
juangaitanv: still open at 060f416; once the `*SQL*` entry is narrowed these cases are the regression guard that keeps it out.

juangaitanv:
Thanks for the quick follow-up! The new commits land well: centralizing the keyword-argument unwrapping into `unwrap_keyword_argument` reads much cleaner, and gating the full-text matching to `database` rules resolves the engine-wide scope concern. I also verified the keyword form end to end on this branch: `cursor.execute(query="..." % uid)` is now flagged and the parameterized keyword call stays clean. Thank you!

I rebuilt and re-scanned at 060f416; a few items from the last round are still open (details in the inline threads):

1. `harness.rs` still fails to compile on unix: line 1133 reads `if executable` inside `#[cfg(unix)]` while the parameter is `_executable` (E0425). Could we fix this first so the build runs on all platforms?
2. the `*SQL*` sanitizer entry (lines 31 and 69 of `sql_injection.ron`) still suppresses real injections: `cursor.execute(SQL_QUERY % user_id)` scans clean on this branch but is flagged on main. Should it narrow to `*sql.SQL(*` / `*psql.SQL(*`?
3. the `*cursor.execute(*f"*` glob still matches f-strings on the params side, so the parameterized `cursor.execute("... %s", [f"user_{uid}"])` is flagged. Could the pattern anchor to the first argument?
4. could we add fixture cases with SQL-named identifiers (e.g. `SQL_QUERY % user_id`) so the sanitizer narrowing in item 2 gets regression cover?

Happy to take another look once these land.

Adityakk9031:
@juangaitanv have look please

Adityakk9031:
@juangaitanv please check the latest commit

Adityakk9031:
@juangaitanv  any feedback

Adityakk9031:
@juangaitanv hey sorry for tagging again just want to check any update

harness.rs
juangaitanv: fixed at 21d013a, the `cfg_attr` form builds clean on macOS.

rules/python/sql_injection.ron
juangaitanv: fixed at 21d013a, verified all three previously suppressed injections now flag.

rules/python/sql_injection.ron
juangaitanv: fixed at 21d013a, the argument-0 anchored rule no longer flags the parameterized f-string case.

rules/python/sql_injection.ron
juangaitanv: fixed at 21d013a together with the line-31 thread.

tests/test_files/python/sql_formatting.py
juangaitanv: added at 21d013a, line 35 is asserted in the strictness test.

juangaitanv (approved):
This lands well! I rebuilt at 21d013a and verified end to end on macOS:

- builds clean on unix, the `cfg_attr` fix is exactly right, and the full test suite passes (195 tests)
- the narrowed sanitizer restores all three previously suppressed injections (`SQL_QUERY % user_id`, `MYSQL_AUDIT` concat, `f"...{SQL_TABLE}..."`)
- anchoring f-string/`.format` detection to argument 0 in `python-sql-raw-format` removes the parameterized f-string false positive while keeping the keyword-argument variants flagged
- corpus diff against main over the python fixtures: 11 false positives removed, 1 new true positive (keyword f-string), no lost detections

One last small ask: could we update the PR description? It still describes the earlier design and does not mention `python-sql-raw-format`, the sanitizer narrowing, the keyword-argument handling, or the harness fix. Not blocking.

Two pre-existing gaps we noticed while verifying (both also miss on main, so out of scope here, happy to file follow-up issues): parenthesized queries like `execute((f"..."))` and implicit string concatenation like `execute("..." f"...")` are not detected.

Adityakk9031:
ok

Ibrahimrahhal (approved):

Adityakk9031:
@juangaitanv and @Ibrahimrahhal done changed the description check it

Adityakk9031:
merge it if it is ok

Adityakk9031:
@juangaitanv and @Ibrahimrahhal  thankyou for review

PR #48, opened 2026-07-15
feat: AST-grounded variable provenance for Python taint analysis

ahmadaburabee:
## Summary

# feat: AST-grounded variable provenance for Python taint analysis

## Problem

Variable provenance — determining which expression assigned a variable's current
value — is resolved by scanning source lines as text: `find_assignment_source`
matches lines starting with `"{var} ="` (`src/scanner/dataflow.rs`), and the
per-file phase-1 tracker records taint from any node whose *text* looks like an
assignment (`src/scanner/scanning_logic.rs`).

Any Python assignment that isn't exactly that shape breaks the analysis in one of
two directions: text that merely looks like an assignment (e.g. example code inside
a docstring) donates false taint, while real assignments the text scan cannot parse
(type-annotated, augmented, chained, tuple, multiline) hide real taint flows. Both
directions are demonstrated on runnable code in **Evidence** below; the
false-positive direction is the pattern discussed in [#40](https://github.com/Corgea/Sighthound/issues/40).

## Change

New module `src/scanner/ast_provenance.rs` resolves provenance from the tree-sitter
AST that Sighthound already builds for every file. For each variable it reports:

- the assignments to that variable in the enclosing function (target, right-hand
side, line, whether the assignment is augmented, and whether the RHS is a
literal collection — i.e. every possible value is developer-controlled), and
- the variable's function-parameter position, which the existing taint logic
already uses when a variable turns out to be a parameter rather than a local.

The resolver reports only structural assignment facts; the existing taint
propagation logic continues to decide whether values are considered tainted.

For Python files, both taint paths first attempt AST resolution and fall back to
the existing text-based logic whenever the AST resolver returns `None`:

- `DataFlowTracer::compute_variable_source` (cross-file analysis) attempts
AST-based resolution before the existing text scan. Results are cached per file.
- Per-file phase-1 source tracking records assignment-borne taint only from real
`assignment` / `augmented_assignment` nodes instead of assignment-looking text,
so docstrings and string literals can no longer donate taint.

The resolver currently supports these assignment forms:

- annotated (`x: str = ...`)
- augmented (`x += ...` — previously skipped by design in the text path)
- chained (`a = b = ...`)
- tuple unpacking (`a, b = ...`)
- multiline / parenthesized right-hand sides
- `for` targets (the loop variable's provenance is the iterable)

It also handles: async `def` scoping, methods and nested functions (a nested
function's locals stay out of the outer scope), literal-collection classification
(a loop over a fixed allowlist — the pattern reported in [#40](https://github.com/Corgea/Sighthound/issues/40) — is classified as
developer-controlled), and f-strings are excluded from literal classification
because their value embeds arbitrary expressions.

## Safety of the change

- Any construct the resolver does not model (e.g. `with ... as`, walrus) returns `None`, and the analysis falls back to the existing text-based implementation
unchanged.
- Non-Python languages always take the existing text path; their behavior is
untouched.
- The resolver returns the existing `VariableSource` enum, so no rule format, API,
or output format changes are required.

## Evidence

Five small Python files scanned with the **default embedded rules** — the only
difference between the two runs is this PR:

| File                 | Code is                                    | `main` (text provenance)        | This PR (AST provenance) |
| -------------------- | ------------------------------------------ | ------------------------------- | ------------------------ |
| `report_export.py`   | safe — parameterized query, constant owner | ❌ **Critical SQL Injection ×2** | ✅ clean                  |
| `feature_flags.py`   | safe — fixed allowlist loop                | ✅ clean                         | ✅ clean                  |
| `account_lookup.py`  | **vulnerable** — annotated assignment      | ❌ sink missed                   | ✅ caught, full trace     |
| `order_archiver.py`  | **vulnerable** — augmented assignment      | ❌ nothing reported              | ✅ caught, full trace     |
| `session_cleanup.py` | **vulnerable** — tuple assignment          | ❌ sink missed                   | ✅ caught, full trace     |

Two excerpts show how lost the text parser gets. The safe file is flagged Critical
with the **docstring** cited as the taint source:

```
● SQL Injection (cwe-89) line 15
source input( ("""Export all reports for the service account.
    ...
        user_id = input("user id: ")     <- documentation, not code
    """)
sink   execute( (export_owner_report)
```

And on the real injection `account_id: str = input(...)`, it extracts the *type
annotation* as the variable name and never reaches the sink:

```
● SQL Injection (cwe-89) line 6          <- the `def` line, not the sink
       var str                            <- the tainted variable is... `str`
```

**Case-by-case before/after on the five demo files (click to expand)**

### 1. False positive: docstring example treated as real code — `report_export.py`

```
 1  """Nightly report export job — SAFE code, no user input anywhere."""
 2
 3  import sqlite3
 4
 5
 6  def export_owner_report(conn):
 7      """Export all reports for the service account.
 8
 9      Example (interactive debugging):
10          user_id = input("user id: ")
11          rows = export_owner_report(conn)
12      """
13      user_id = "report-service"
14      cursor = conn.cursor()
15      cursor.execute("SELECT * FROM reports WHERE owner = ?", (user_id,))
16      return cursor.fetchall()
```

Before: 2 Critical findings at line 15 (the `execute` sink), source attributed
to the `input(` example on line 10 inside the docstring. After: clean — a
docstring is a string node, not an assignment. This is the false-positive
pattern discussed in [#40](https://github.com/Corgea/Sighthound/issues/40): a
provably developer-controlled value indistinguishable from attacker input
because provenance was guessed from text.

### 2. False negative: type annotation hides an injection — `account_lookup.py`

```
 1  """Account lookup CLI — VULNERABLE: user input reaches SQL via an annotated assignment."""
 2
 3  import sqlite3
 4
 5
 6  def lookup_account(conn):
 7      account_id: str = input("account id: ")
 8      query = f"SELECT * FROM accounts WHERE id = '{account_id}'"
 9      cursor = conn.cursor()
10      cursor.execute(query)
11      return cursor.fetchone()
```

Before: the `: str` annotation on line 7 breaks the `account_id =` text match —
the sink on line 10 is never reported; the one emitted finding names `var str`
at the `def` line (line 6). After:

```
● SQL Injection (cwe-89) line 10
source input( (Propagated from account_id via: query = f"SELECT * FROM accounts WHERE id = '{account_id}'")
sink   execute( (lookup_account)
       var query
```

### 3. False negative: `+=` skipped by design — `order_archiver.py`

```
 1  """Order archive CLI — VULNERABLE: user input reaches SQL via an augmented assignment."""
 2
 3  import sqlite3
 4
 5
 6  def archive_orders(conn):
 7      query = "DELETE FROM orders WHERE ref IN ("
 8      query += input("order refs (comma separated): ")
 9      query += ")"
10      cursor = conn.cursor()
11      cursor.execute(query)
12      conn.commit()
```

Before: zero findings — the text path skips augmented assignments
(`src/scanner/dataflow.rs`: "augmented assignments are intentionally not
matched"), so the `+=` on line 8 is invisible. After: caught at line 11 (the
`execute` sink) with the line-8 `+=` cited as the source. Structurally, `+=` is
just an `augmented_assignment` node.

### 4. False negative: tuple unpacking taints the wrong variable — `session_cleanup.py`

```
 1  """Session cleanup CLI — VULNERABLE: user input reaches SQL via a tuple assignment."""
 2
 3  import sqlite3
 4
 5
 6  def purge_session(conn):
 7      session_id, keep_days = input("session id: "), 30
 8      query = f"DELETE FROM sessions WHERE id = '{session_id}'"
 9      cursor = conn.cursor()
10      cursor.execute(query)
11      conn.commit()
12      return keep_days
```

Before: the finding names `var keep_days` — the constant `30` on line 7 — while
`session_id` (the actual attacker input) is never connected; the sink is missed.
After: `input( → session_id → query → execute(` reported at line 10.

### 5. Precision guard: the fixed allowlist loop pattern from [#40](https://github.com/Corgea/Sighthound/issues/40) — `feature_flags.py`

```
 1  """Feature flag seeding — SAFE: values come from a fixed allowlist (the issue #40 shape)."""
 2
 3  import sqlite3
 4
 5
 6  def seed_flags(conn):
 7      cursor = conn.cursor()
 8      for flag in ["beta", "dark_mode", "canary"]:
 9          cursor.execute("INSERT INTO flags (name) VALUES (?)", (flag,))
10      conn.commit()
```

Clean before, clean after — the added recall costs no noise, and the resolver now
*proves* this safe (literal-collection classification) rather than accidentally
not reporting it.

## Tests

- 8 new strictness fixtures (`tests/strictness/ast_provenance.rs`): **4 fail on**
`main` — the three missed sinks and the docstring false positive above — and
all 8 pass on this branch.
- 8 new unit tests pin the resolver's grammar handling (type-hinted/default/splat
parameters, chained/tuple targets, nested-scope isolation, f-string exclusion
from literal collections, async `def`).
- Full suite green across all harnesses (`cargo test`): unit, integration,
end-to-end, strictness, acceptance. No pre-existing test was modified.

Reproduce: check out this branch vs `main` and run  
`cargo run --quiet --bin sighthound -- <demo dir>` on the snippets above.

## Related issue

[#40](https://github.com/Corgea/Sighthound/issues/40)

## Checklist

- [ ] `cargo build --release` succeeds
- [ ] Ran any test harness relevant to this change (note results; suite under repair)
- [ ] Updated docs/rules where relevant


yhoztak:
@ahmadaburabee thank you for this contribution. This is great idea and changes look very solid. Reviewing it through now.

src/scanner/ast_provenance.rs
```diff
@@ -0,0 +1,587 @@
...
+            out.push(AssignmentFact {
+                target: name,
+                rhs: rhs.clone(),
+                line,
+                augmented,
+                literal_collection,
+            });
+        }
+    }
+}
```
yhoztak, line 428: `push_target_facts` computes a single rhs from the whole value node and clones it onto every target, so in a tuple assignment each left identifier inherits the entire right-hand side.

```
cmd, log_name = input(), "app.log"
os.system(log_name)   # falsely flagged
```

The existing tuple test only sinks cmd, so it can't catch this.

Suggest pairing tuple targets positionally with the corresponding RHS element (zip left pattern_list/tuple_pattern children against a expression_list/tuple/list RHS of equal arity), and falling back to today's whole-RHS behavior on arity mismatch (e.g. a, b = returns_tuple(), starred targets) so we don't lose taint. Please also add a negative test that sinks the safe sibling and asserts no finding.

One example of fix may be to have 

```
/// Pair each assignment target with the value node it actually receives.
/// Tuple/list targets are matched positionally against a tuple/list RHS of
/// equal arity; any other shape (arity mismatch, call returning a tuple,
/// starred target) conservatively attributes the whole RHS to each name.
fn target_value_pairs<'a>(
    target: Node<'a>,
    value: Node<'a>,
    source: &[u8],
) -> Vec<(String, Node<'a>)> {
    match target.kind() {
        "pattern_list" | "tuple_pattern" => {
            let names = tuple_element_nodes(target);
            if let Some(values) = matching_rhs_elements(value, names.len()) {
                return names
                    .iter()
                    .zip(values)
                    .flat_map(|(name_node, val)| {
                        target_identifiers(*name_node, source)
                            .into_iter()
                            .map(move |n| (n, val))
                    })
                    .collect();
            }
            // Arity we can't line up: keep today's conservative behavior.
            target_identifiers(target, source)
                .into_iter()
                .map(|n| (n, value))
                .collect()
        }
        _ => target_identifiers(target, source)
            .into_iter()
            .map(|n| (n, value))
            .collect(),
    }
}
/// Named element nodes of a tuple/list literal or pattern, ignoring commas.
fn tuple_element_nodes(node: Node) -> Vec<Node> {
    let mut cursor = node.walk();
    node.named_children(&mut cursor).filter(|c| c.kind() != "comment").collect()
}
/// `Some(elements)` only when `value` is a tuple/list literal with exactly
/// `arity` elements; `None` signals "cannot line up, be conservative".
fn matching_rhs_elements(value: Node, arity: usize) -> Option<Vec<Node>> {
    match value.kind() {
        "expression_list" | "tuple" | "list" => {
            let elems = tuple_element_nodes(value);
            (elems.len() == arity).then_some(elems)
        }
        _ => None,
    }
}
```

Then push_target_facts computes rhs/literal_collection per value node instead of once:

```
fn push_target_facts(
    statement: Node,
    targets: &[Node],
    value: Node,
    augmented: bool,
    source: &[u8],
    out: &mut Vec<AssignmentFact>,
) {
    let line = statement.start_position().row + 1;
    for target in targets {
        for (name, value_node) in target_value_pairs(*target, value, source) {
            out.push(AssignmentFact {
                target: name,
                rhs: normalized_text(value_node, source),
                line,
                augmented,
                literal_collection: is_literal_collection(value_node),
            });
        }
    }
}
```

src/scanner/ast_provenance.rs
```diff
@@ -0,0 +1,587 @@
...
+    if node.kind() == "function_definition" {
+        record_function(node, source, functions);
+    }
+    let mut cursor = node.walk();
+    for child in node.children(&mut cursor) {
+        collect_functions(child, source, functions);
+    }
+}
+
+fn record_function(
```
yhoztak, line 227: Functions are stored in FileFacts.functions: BTreeMap<String, FunctionFacts> keyed on the short name, and record_function keeps only the first definition (if functions.contains_key(&name) { return; }). Since resolve_variable is called with the sink's bare function context (e.g. "run"), a sink can resolve against a different, same-named function:

```
lass Unsafe:
    def run(self): cmd = input()
class Safe:
    def run(self):
        cmd = "uptime"
        os.system(cmd)   # can resolve to Unsafe.run's cmd
```

To be clear on severity: this is new code, but not a new regression vs main — the text fallback (extract_function_body) already matches the first def run(, so both paths pick the same wrong function and the output is unchanged. The concern is that we're introducing a fresh module that carries the limitation forward, and returning Some(...) here shadows the fallback, so it's the right time to do better.

Minimal safe fix: on a duplicate name, mark the entry ambiguous and have resolve_variable return None so the caller falls back (no worse than today, but drops the confidently-wrong result). Proper fix (can be a follow-up): key by source span / qualified name (Class.method) and resolve using the sink's enclosing function_definition node instead of the bare name.

Not blocking merge on its own, but worth a decision here rather than silently inheriting it.

ahmadaburabee:
Thanks for the careful review, @yhoztak — both points were spot on. I've addressed them; summary and before/after below. Full suite stays green (15 resolver unit tests, 33 strictness fixtures), and I added tests that lock in each fix.

## 1. Tuple unpacking — positional pairing (fixed)

You were right that `push_target_facts` attributed the whole right-hand side to every target, so a safe sibling in a tuple unpack inherited a tainted element's value. I took your positional-pairing idea and implemented it as `target_value_pairs` / `positional_pairs`: when both sides are sequences of equal length and every target is a plain identifier, each name is matched to *its own* value element; anything else (starred target, nesting, subscript element, a non-sequence RHS like `a, b = f()`, or an arity mismatch) falls back conservatively to whole-value attribution, so no taint is ever dropped. Pairing applies to assignments only — `for` targets are deliberately excluded, since `for a, b in x, y:` unpacks *each* element of the iterated tuple into `(a, b)`, so every loop target keeps the whole iterable (regression test: `for_loop_targets_keep_the_whole_iterable`).

The effect is actually stronger than just silencing the false positive — on taint rules that treat `input()` as a command-injection source (e.g. `tests/strictness/fixtures/command_injection_taint.ron`), the old behavior was flagging the **wrong** variable:

```
1  import os
2
3
4  def run():
5      cmd, log_name = input("c: "), "app.log"
6      os.system(log_name)      # SAFE  — log_name is the constant "app.log"
7      os.system(cmd)           # VULNERABLE — cmd came from input()
```

**Before (main):**
```
line 6 -> flags 'log_name'      ❌ false positive on the safe sibling
(line 7 os.system(cmd) is never reported — the real injection is missed)
```

**After (this PR):**
```
line 7 -> flags 'cmd'           ✅ the real injection
(line 6 os.system(log_name) is correctly left clean)
```

So positional pairing both removes the false positive on `log_name` *and* surfaces the true positive on `cmd` that `main` was missing. Regression tests added: `tuple_safe_sibling_is_not_flagged` (safe sibling stays clean) and the existing `tuple_taint_assignment_is_detected` (tainted element still reported), plus unit tests for the pairing and the conservative fallback.

## 2. Same-named functions — ambiguity guard

You're right that `record_function` keeps only the first definition of a bare name, so a sink could resolve against a same-named function in another class. I hardened the resolver: any bare name defined more than once is now recorded as **ambiguous** and resolves to `None`, so the resolver defers instead of guessing (unit test: `duplicate_function_names_are_ambiguous_and_yield_none`).

Your concern wasn't hypothetical — I confirmed it with a before/after on this branch's cross-file path. Take two same-named functions where the tainted one uses a shape only the AST can see:

```python
import os

class Unsafe:
    def run(self):
        cmd: str = input()      # line 5 — annotated, invisible to the text scan

class Safe:
    def run(self):
        cmd = "uptime"
        os.system(cmd)
```

Without the guard, the resolver resolves the bare name `run` to its first definition and attributes `Unsafe.run`'s `input()` to `Safe.run`'s sink, emitting a misattributed cross-file flow (to reproduce: scan alongside at least one other `.py` file — the cross-file phase is skipped for single-file scans):

```
without guard: [COMPUTE_VARIABLE_SOURCE] AST: tainted assignment 'input()' at line 5
               [CROSS_FILE_NEW] Enhanced analysis complete. Found 1 verified flows   ❌ wrong `run`
with guard:    [COMPUTE_VARIABLE_SOURCE] Variable 'cmd' source unknown, defaulting to parameter 0
               [CROSS_FILE_NEW] Enhanced analysis complete. Found 0 verified flows   ✅ parity with main
```

So the guard is load-bearing, not belt-and-braces — exactly the failure you predicted.

Two honest caveats worth flagging:

1. Your exact example (plain `cmd = input()`) reproduces **identically on `main`** — 2 findings on both branches, byte-for-byte. That false positive comes from the pre-existing per-file tracker (`VariableFlowTracker` keys taint by bare function name: `info.source_function == function`), and the new resolver isn't even invoked there — single-file scans skip the cross-file phase entirely. Debug logs on `main` show the mechanism: `[RECORD_TAINT] ... 'cmd' ... in function 'run'` (from `Unsafe.run`) followed by `[CHECK_TAINT] Variable 'cmd' is tainted (function match or global)` at `Safe.run`'s sink. Not a regression introduced here, and the resolver no longer contributes to it.

2. For assignment shapes only the AST path can parse, this PR's better extraction can *newly expose* that tracker collision: on the annotated variant above, `main` reports 0 findings — only because its text scan misparses `cmd: str = input()` and taints the name `str` — while this branch correctly extracts `cmd` and so surfaces 2 tracker findings on `Safe.run`. The hazard is pre-existing tracker design, but I want to be upfront that improved extraction widens its exposure on same-named functions.

Both caveats point at the same root cause you identified: bare-name keying in the shared tracker. The proper end-to-end fix is qualified-name / source-span resolution there, which I'd propose as a focused follow-up rather than expand this PR's scope — and caveat 2 is a good argument for making it a *fast* follow-up. Happy to pick that up next if it fits the roadmap.

## Tests

- Resolver unit tests: 15 (added tuple pairing, conservative fallback, for-loop pairing exclusion, ambiguity guard).
- Strictness fixtures: 33 (added the safe-sibling regression guard).
- Full `cargo test` green across unit, integration, end-to-end, strictness, and acceptance. No pre-existing test modified.

src/scanner/dataflow.rs
```diff
@@ -394,6 +407,67 @@ impl DataFlowTracer {
...
+            if let ValueSourceClassification::Tainted(pattern) =
+                self.classify_value_source(&fact.rhs, rule_deduplicator)
+            {
+                log::debug!(
+                    "[COMPUTE_VARIABLE_SOURCE] AST: tainted assignment '{}' at line {}",
+                    fact.rhs,
+                    fact.line
+                );
+                return Some(VariableSource::DirectTaintSource { pattern, line: fact.line });
+            }
```
juangaitanv, line 522: should this only consider assignments that precede the sink line? `cmd = "safe"; os.system(cmd); cmd = input()` looks like it would mark the earlier sink tainted.

src/scanner/ast_provenance.rs
```diff
@@ -0,0 +1,736 @@
...
+/// yet modeled. Tuple targets conservatively attribute the whole RHS to each
+/// plain-identifier element.
+fn target_identifiers(node: Node, source: &[u8]) -> Vec<String> {
+    match node.kind() {
+        "identifier" => vec![get_node_text_slice(&node, source).to_string()],
+        "pattern_list" | "tuple_pattern" => {
+            let mut cursor = node.walk();
+            node.named_children(&mut cursor)
+                .filter(|child| child.kind() == "identifier")
+                .map(|child| get_node_text_slice(&child, source).to_string())
```
juangaitanv: should whole-value fallback recurse into starred and nested targets? `head, *tail = "safe", input()` records the tainted whole RHS only against `head`, creating a false positive for `head` while dropping taint from `tail`.

src/scanner/dataflow.rs
```diff
@@ -360,6 +364,15 @@ impl DataFlowTracer {
...
         );
 
+        // AST-grounded provenance first (Python). Any shape the resolver does not
+        // model yields `None`, so the text path below keeps handling it exactly as
+        // before — other languages always take the text path.
+        if let Some(source) =
+            self.ast_variable_source(file_path, function_name, variable_name, rule_deduplicator)
+        {
+            return source;
+        }
```
juangaitanv: when AST resolution returns `None` because `function_name` is ambiguous, should we skip the text fallback? that path still keys off the first matching definition and can reintroduce the cross-definition mix-up.

src/scanner/scanning_logic.rs
```diff
@@ -963,10 +964,114 @@ impl ScanningLogic {
...
+                    ctx,
+                    flow_tracker,
+                );
+            }
+            return;
+        }
+
+        let facts = match node.kind() {
+            "assignment" | "augmented_assignment" => {
+                ast_provenance::assignment_facts_for_node(*node, ctx.source)
```
juangaitanv, line 1028: should Python assignments with no structural facts fall back to the text tracker? `self.cmd = input(); cmd = self.cmd; os.system(cmd)` loses the source that the previous path recorded.

src/scanner/dataflow.rs
```diff
@@ -394,6 +407,67 @@ impl DataFlowTracer {
...
+
+        let first = assignments.iter().find(|fact| !fact.augmented)?;
+        if first.literal_collection {
+            log::debug!("[COMPUTE_VARIABLE_SOURCE] AST: literal collection at line {}", first.line);
+            return Some(VariableSource::KnownSafe {
+                reason: "value drawn from a literal collection".to_string(),
+                line: first.line,
+            });
+        }
```
juangaitanv: could we return `KnownSafe` for a literal collection only when every assignment fact is literal? a later `cfg["cmd"] = user_supplied` can otherwise be skipped and the sink treated as definitely safe.

ahmadaburabee:
Thanks @juangaitanv ,all five were real; fixed in the latest commit: the resolver is now sink-line-aware (loop-carried assignments still reach), starred targets pair around the star and the fallback never drops names, ambiguous function names return Unknown instead of falling back to a first-definition text guess, factless assignments (attribute targets) fall back to the text tracker, that one was a real regression vs main , and `KnownSafe` requires every reaching write to be literal. 8 new unit tests + 5 strictness fixtures; all pre-existing fixtures green.

juangaitanv:
@ahmadaburabee thank you for addressing the comments! could you review the CI /harness failures?

ahmadaburabee:
@juangaitanv Done 👍

src/scanner/dataflow.rs
```diff
@@ -394,6 +432,95 @@ impl DataFlowTracer {
...
+        // fact against `cfg`) can replace the literal contents, so it must
+        // defeat the definitely-safe verdict rather than be skipped.
+        if first.literal_collection && assignments.iter().all(|fact| fact.literal_value) {
+            log::debug!("[COMPUTE_VARIABLE_SOURCE] AST: literal collection at line {}", first.line);
+            return Some(VariableSource::KnownSafe {
+                reason: "value drawn from a literal collection".to_string(),
+                line: first.line,
+            });
+        }
+        Some(self.classify_assignment_rhs(&first.rhs, first.line, rule_deduplicator))
```
juangaitanv: should the latest reaching write provide provenance instead of the first? `cmd: str = "safe"; cmd = read_command(); os.system(cmd)` classifies from the initializer and returns `KnownSafe`, so the flow main traced through `read_command()` is no longer reported.

src/scanner/ast_provenance.rs
```diff
@@ -0,0 +1,988 @@
...
+/// Named elements of a sequence node on either side of an assignment: a target
+/// list/tuple pattern, or a value tuple / bare `a, b` expression list.
+fn sequence_elements(node: Node) -> Option<Vec<Node>> {
+    match node.kind() {
+        "pattern_list" | "tuple_pattern" | "tuple" | "expression_list" => {
+            let mut cursor = node.walk();
+            Some(node.named_children(&mut cursor).filter(|c| c.kind() != "comment").collect())
+        }
+        _ => None,
+    }
```
juangaitanv, line 545: could `sequence_elements` accept `list` values too? `cmd, log_name = [input(), "app.log"]` falls back to whole-value attribution and flags `os.system(log_name)`.

src/scanner/dataflow.rs
juangaitanv: selecting the first plain assignment can hide a later reaching write: `cmd: str = "safe"; cmd = read_command(); os.system(cmd)` returns `KnownSafe` and skips the call. Classifying the latest reaching write would keep the flow traceable.

src/scanner/ast_provenance.rs
juangaitanv, line 545: list destructuring is not paired positionally. `cmd, log_name = [input(), "app.log"]` falls back to assigning the full RHS to both names, so `os.system(log_name)` can be reported as tainted. Handling `list` values here would preserve Python bindings.

src/scanner/dataflow.rs
```diff
@@ -394,6 +432,95 @@ impl DataFlowTracer {
...
+        // itself literal — a later `cfg["cmd"] = user_supplied` (recorded as a
+        // fact against `cfg`) can replace the literal contents, so it must
+        // defeat the definitely-safe verdict rather than be skipped.
+        if first.literal_collection && assignments.iter().all(|fact| fact.literal_value) {
+            log::debug!("[COMPUTE_VARIABLE_SOURCE] AST: literal collection at line {}", first.line);
+            return Some(VariableSource::KnownSafe {
+                reason: "value drawn from a literal collection".to_string(),
+                line: first.line,
+            });
+        }
```
juangaitanv: a non-empty literal collection can remain `KnownSafe` after a tainting mutator. `parts = ["prefix"]; parts.append(input()); os.system(" ".join(parts))` suppresses the flow. Reaching collection mutations need to prevent this safe classification.

ahmadaburabee:
Thanks @juangaitanv , all three fixed in the latest commit: provenance now comes from the latest reaching write (not the first), list-literal RHS is paired positionally like tuples, and collection mutations are recorded as facts so a non-empty literal collection can't stay KnownSafe after a tainting mutator. Added 11 tests covering each. The list-sibling case was a real false positive end-to-end; the other two were correctness fixes in the cross-file classifier.

src/scanner/dataflow.rs
```diff
@@ -394,6 +432,105 @@ impl DataFlowTracer {
...
+            );
+            return Some(VariableSource::KnownSafe {
+                reason: "value drawn from a literal collection".to_string(),
+                line: initializer.line,
+            });
+        }
+        // Provenance is the latest reaching plain write: at the sink the variable
+        // holds whatever was assigned most recently, not its first initializer.
+        let latest = assignments.iter().rev().find(|fact| !fact.augmented).unwrap_or(initializer);
+        Some(self.classify_assignment_rhs(&latest.rhs, latest.line, rule_deduplicator))
```
juangaitanv: should we trace sibling-branch reaching writes independently so `cmd = "ls"` does not mask `cmd = fetch_remote()` at the sink?

src/scanner/dataflow.rs
```diff
@@ -394,6 +432,105 @@ impl DataFlowTracer {
...
+                return Some(VariableSource::DirectTaintSource { pattern, line: fact.line });
+            }
+        }
+
+        // A literal collection proves safety only when it was initialized as a
+        // literal AND every reaching write is itself literal — a later
+        // `cfg["cmd"] = user_supplied` or `parts.append(input())` (recorded as a
+        // non-literal fact against the collection) must defeat the
+        // definitely-safe verdict rather than be skipped.
+        let initializer = assignments.iter().find(|fact| !fact.augmented)?;
```
juangaitanv: should we return Unknown or trace augmented RHS facts before fallback so below-sink assignments cannot supply provenance?

yhoztak:
@ahmadaburabee thank you for updating the PR, it's getting close and here are some more findings that should be addressed before merging this PR. 

1. Safe SQL falsely reported. This passes on main but fails here.

```
mp=$(mktemp -d)
cat > "$tmp/safe.py" <<'PY'
def get_user(cursor, user_id):
    cursor.execute(
        "SELECT * FROM users WHERE id = %s",
        (user_id,),
    )
PY
cargo run --quiet --bin sighthound -- \
  "$tmp" python rules/python/sql_injection.ron \
  --use-file-rules --simple-analysis --output-format json
```


2. Added this to `dataflow.rs` for test case and it failed:
```
#[test]
fn indirect_augmented_rhs_supplies_provenance() {
    let tracer = DataFlowTracer::new();
    let dedup = TaintRuleDeduplicator::new(&[]);
    let facts = vec![
        fact("cmd", "\"safe\"", 1, false, false, true),
        fact("cmd", "user_value", 2, true, false, false),
    ];

    assert!(matches!(
        tracer.classify_ast_assignments(&facts, &dedup),
        Some(VariableSource::LocalAssignment {
            source_expression,
            ..
        }) if source_expression == "user_value"
    ));
}
```

ahmadaburabee:
@yhoztak  Both valid, thanks. The SQL FP is the old python-sql-raw-001 %-glob matching the %s placeholder — that's issue #35, fixed on main by PR #44, and this branch was just based on an older main. Merged origin/main in (clean, no conflicts); the parameterized query now reports 0 findings and PR #44's SQL regression tests pass. The augmented-RHS case is fixed by the classifier rewrite , cmd += user_value now yields LocalAssignment { source_expression: "user_value" } instead of KnownSafe, and I added your test to lock it.

yhoztak (approved):
LGTM
Thank you for your contribution!

PR #51, opened 2026-07-18
feat(ruby): AST-based command injection false positive suppression

Adityakk9031:
## Problem
The Ruby security rules suffered from false-positives under safe multi-argument executions (e.g., using array literals or splatted array arguments inside command execution methods). These multi-argument formats are securely processed by the OS directly, bypassing the shell. Additionally, there were minor Clippy and formatting warnings that prevented quality harness runs from finishing cleanly.

## Solution
1. **AST-Based Suppression**: Added a custom AST-based predicate (`ruby_unsafe_command_injection`) to differentiate safe executions (e.g., multi-argument arrays) from unsafe ones (e.g., interpolated strings, single variables) during command injection scans.
2. **Scanner Integration**: Integrated the AST inspection conditions directly into the conditions evaluation logic (`src/scanner/conditions.rs`) and taint scanner logic (`src/scanner/scanning_logic.rs`).
3. **Rigorous Testing**: Added extensive safe and unsafe Ruby test fixtures along with integration tests (`tests/strictness/language_coverage.rs`) to ensure false-positive suppression matches correctness requirements.
4. **Code Quality & Clippy Fixes**:
   - Resolved `clippy::items-after-test-module` in `src/scanner/output.rs` by moving the test module to the end of the file.
   - Fixed `is_empty` and unnecessary `return` statements in `tests/unit/django_xss_prevention_tests.rs`.
   - Replaced redundant clones with slice references in `harness.rs`.
   - Verified that `cargo test`, `cargo clippy`, and `./target/debug/harness-run.exe check` all pass with zero warnings or errors.

## Changes
- `rules/ruby/command_injection.ron`: Integrated custom AST conditions.
- `src/scanner/conditions.rs`: Implemented AST predicate logic and evaluation helpers.
- `src/scanner/scanning_logic.rs`: Incorporated conditions checking during post-taint analysis.
- `src/scanner/output.rs`: Relocated the SARIF tests module to resolve Clippy lints.
- `tests/unit/django_xss_prevention_tests.rs`: Fixed styling lints.
- `harness.rs`: Fixed reference-to-slice clone warning.
- `tests/test_files/strictness_languages/ruby/`: Updated `safe.rb` and `unsafe.rb` fixtures.
- `tests/strictness/language_coverage.rs`: Added the `ruby_rules_taint_and_search_validation` regression test.


Adityakk9031:
@asadeddin @juangaitanv have a look

src/scanner/conditions.rs
```diff
@@ -219,3 +228,88 @@ pub fn check_has_sibling_pattern_condition(
...
+    language_support: &dyn LanguageSupport,
+) -> bool {
+    if language_support.name() != "ruby" {
+        return true;
+    }
+    if let Some(args_node) = language_support.get_arguments_node(node) {
+        let count = args_node.named_child_count();
+        if count > 1 {
+            return false; // Safe: multi-argument form bypassed the shell
+        }
```
juangaitanv: could we avoid treating every multi-argument call as shell-free? `system("sh", "-c", params[:cmd])` explicitly executes the tainted value through a shell, and options or env hashes can leave a single command string shell-interpreted. Suppress only when the call shape proves argv-style execution.

tests/strictness/language_coverage.rs
```diff
@@ -274,3 +274,59 @@ fn finding_key_set(findings: &[Value]) -> BTreeSet<(String, u64, String, String)
...
+
+    // Unsafe patterns must trigger findings
+    assert!(
+        unsafe_findings.len() >= 6,
+        "unsafe.rb should trigger at least 6 findings, got {}: {:?}",
+        unsafe_findings.len(),
+        unsafe_findings
+            .iter()
+            .map(|f| (f.line, f.finding_type.as_str(), f.snippet.as_str()))
+            .collect::<Vec<_>>()
```
juangaitanv, line 547: could we assert the taint-only sink lines as well? The six direct interpolation cases already satisfy this aggregate count, so the test can pass even if condition enforcement no longer emits findings for `system(cmd)` and `system(*cmd)`.

Adityakk9031:
@juangaitanv have a look

src/scanner/conditions.rs
```diff
@@ -219,3 +228,154 @@ pub fn check_has_sibling_pattern_condition(
...
+            );
+            if is_shell {
+                for arg_node in cmd_args.iter().skip(1) {
+                    let arg_text = clean_ruby_string(get_node_text(arg_node, source));
+                    if arg_text == "-c" || arg_text == "/c" || arg_text == "/C" {
+                        return true; // Explicit shell execution: UNSAFE
+                    }
+                }
+            }
+            return false; // Safe: multi-argument form bypassed the shell
```
juangaitanv: should we distinguish callee signatures before treating multi-argument calls as safe, given `IO.popen(cmd, mode)` and `Open3.pipeline` still execute positional command strings through a shell?

juangaitanv: shouldn't the shell check recognize executable paths such as `/bin/sh` before suppressing `system("/bin/sh", "-c", params[:cmd])`?

Adityakk9031:
@juangaitanv and @Ibrahimrahhal  hey please check this

src/scanner/conditions.rs
```diff
@@ -24,15 +24,24 @@ pub fn check_single_condition(
...
         "has_ancestor" => check_has_ancestor_condition(node, condition),
         "argument_not_sanitized" => {
             check_argument_not_sanitized_condition(node, source, condition, language_support)
         }
         "has_sibling_pattern" => check_has_sibling_pattern_condition(node, source, condition),
-        _ => false,
+        "ruby_unsafe_command_injection" => {
+            check_ruby_unsafe_command_injection(node, source, language_support)
+        }
+        _ => evaluate_field_condition(node, source, condition),
```
juangaitanv: should we limit generic fallback evaluation to conditions designed for it, given `originalUrl` can satisfy the JavaScript `not_contains "origin"` context check?

src/scanner/conditions.rs
```diff
@@ -219,3 +228,239 @@ pub fn check_has_sibling_pattern_condition(
...
+                        return true; // Explicit shell execution: UNSAFE
+                    }
+                }
+            }
+            return false; // Safe: multi-argument form bypassed the shell
+        }
+        if count == 1 {
+            let arg = &cmd_args[0];
+            if arg.kind() == "array" {
+                return false; // Safe: single array literal argument bypassed the shell
```
juangaitanv: should we require a static executable before classifying multi-argument and array command calls as safe?

src/scanner/conditions.rs
```diff
@@ -219,3 +228,239 @@ pub fn check_has_sibling_pattern_condition(
...
+                        if child.kind() == "array" {
+                            has_array_child = true;
+                            break;
+                        }
+                    }
+                }
+                if has_array_child {
+                    return false; // Safe: splatted array literal bypassed the shell
+                }
+            }
```
juangaitanv: should we apply the shell-and-command-flag check to array and splatted-array arguments before classifying them as safe?

src/scanner/conditions.rs
```diff
@@ -219,3 +228,239 @@ pub fn check_has_sibling_pattern_condition(
...
+                    for i in 0..arg.named_child_count() {
+                        if let Some(child) = arg.named_child(i as u32) {
+                            if child.kind() == "array" {
+                                return false; // Safe: splatted array argument
+                            }
+                        }
+                    }
+                }
+            }
+            return true; // Single-string command passed to IO.popen is UNSAFE
```
juangaitanv: should we normalize namespace-qualified callees before matching `IO.popen` and `Open3.pipeline*`?

src/scanner/conditions.rs
```diff
@@ -219,3 +228,239 @@ pub fn check_has_sibling_pattern_condition(
...
+            let first_text = clean_ruby_string(get_node_text(&cmd_args[0], source));
+            if is_shell_executable(&first_text) {
+                for arg_node in cmd_args.iter().skip(1) {
+                    let arg_text = clean_ruby_string(get_node_text(arg_node, source));
+                    if arg_text == "-c" || arg_text == "/c" || arg_text == "/C" || arg_text == "-C"
+                    {
+                        return true; // Explicit shell execution: UNSAFE
+                    }
+                }
+            }
```
juangaitanv: should we recognize `-Command` and encoded-command flags for `powershell` and `pwsh`?

juangaitanv:
thank you for sticking with this through several rounds 🙏  most of the new comments come down to one question:
should unmodeled call shapes fail closed rather than open, since a miss here becomes a silent false negative?

settling that once likely clears comments 2–5 together. comment 1 is separate: the new condition fallback activates previously-dormant conditions in other languages' rules, which may deserve its own scoping decision.

resolved the two earlier threads, thank you for addressing those.

Adityakk9031:
Thanks @juangaitanv  I've updated the PR based on your feedback:

Adopted a fail-closed approach: dynamic executables or array calls invoking shells with -c/-Command/-EncodedCommand (including PowerShell/pwsh) are flagged as unsafe.
Normalized namespace callees (::IO.popen, Open3::pipeline).
Scoped condition fallbacks to pattern and context fields so dormant rules aren't activated.
Added test fixtures for all new cases, and make ci / clippy pass cleanly.

src/scanner/scanning_logic.rs
```diff
@@ -1006,6 +1006,16 @@ impl ScanningLogic {
         if TaintExpressionUtils::expression_has_sanitizer(rule, site.node_text) {
             return;
         }
+        if let Some(conditions) = &rule.conditions {
+            if !crate::scanner::conditions::check_ast_conditions(
```
juangaitanv: could the taint-path condition gate be scoped to the ruby command-injection condition type, given ~50 JavaScript findings (CWE-346, CWE-922, DOM XSS) disappear on the current fixtures?

src/scanner/scanning_logic.rs
```diff
@@ -1066,6 +1077,16 @@ impl ScanningLogic {
                 );
                 continue; // Skip this finding as it's sanitized
             }
+            if let Some(conditions) = &rule.conditions {
+                if !crate::scanner::conditions::check_ast_conditions(
```
juangaitanv: should the condition check run against every node the taint walk visits, given object literals like `cwe_922_config_store_test.js:108` reach `check_not_literal_condition`, get `None` args, and drop the finding?

src/scanner/scanning_logic.rs
```diff
@@ -1124,12 +1145,6 @@ impl ScanningLogic {
         let Some(sink_pattern) = ctx.rule_deduplicator.matches_sink_pattern(&node_text) else {
```
juangaitanv, line 1373: was removing the `[SINK_ANALYSIS] Found sink` debug log intended?

src/common.rs
```diff
@@ -563,7 +572,15 @@ impl CommonUtils {
         Some(
             args_str
                 .split(',')
-                .map(|s| s.trim().to_string())
+                .map(|s| {
+                    let mut s_trimmed = s.trim();
+                    if s_trimmed.starts_with('*') {
```
juangaitanv: could the leading `*`/`&` stripping and the `extract_simple_variables` decomposition be confined to ruby, given `cursor.execute("... %s", [firm_id])` now surfaces `firm_id` and re-flags the issue-#35 safe fixture?

src/scanner/conditions.rs
```diff
@@ -24,15 +24,30 @@ pub fn check_single_condition(
 ) -> bool {
     match condition.condition_type.as_deref().unwrap_or("") {
         "has_argument" => check_has_argument_condition(node, source, condition, language_support),
-        "in_context" => check_in_context_condition(node, condition),
+        "in_context" => {
+            if condition.not_in.is_some() {
+                check_in_context_condition(node, condition)
+            } else {
+                evaluate_field_condition(node, source, condition)
```
juangaitanv: should the `in_context` route to `evaluate_field_condition` be scoped more tightly, given `js-missing-origin-validation-taint-001` has no `not_in` and its `not_contains "origin"` check now fires on strings like `originalUrl`?

src/scanner/conditions.rs
```diff
@@ -24,15 +24,30 @@ pub fn check_single_condition(
...
         "argument_not_sanitized" => {
             check_argument_not_sanitized_condition(node, source, condition, language_support)
         }
         "has_sibling_pattern" => check_has_sibling_pattern_condition(node, source, condition),
-        _ => false,
+        "ruby_unsafe_command_injection" => {
+            check_ruby_unsafe_command_injection(node, source, language_support)
+        }
+        _ => {
+            if condition.field == "pattern" || condition.field == "context" {
```
juangaitanv: could the generic fallback be limited to condition types designed for it rather than any condition whose field is `pattern` or `context`, given the ~30 dormant JS conditions that now activate?

src/scanner/conditions.rs
```diff
@@ -219,3 +234,287 @@ pub fn check_has_sibling_pattern_condition(
...
+        val = val[1..val.len() - 1].to_string();
+    }
+    if val.starts_with(':') {
+        val = val[1..].to_string();
+    }
+    val.trim().to_string()
+}
+
+/// Extract shell name from a path or executable string, e.g. "/bin/sh" -> "sh", "C:\Windows\cmd.exe" -> "cmd"
+fn extract_shell_name(text: &str) -> Option<String> {
```
juangaitanv, line 276: could the shell-name and flag helpers get direct unit tests, e.g. `extract_shell_name`, `is_shell_command_flag`, `clean_ruby_string`?

src/scanner/conditions.rs
```diff
@@ -219,3 +234,287 @@ pub fn check_has_sibling_pattern_condition(
...
+    }
+
+    let first_kind = elements[0].kind();
+    matches!(first_kind, "string" | "string_literal" | "simple_symbol" | "symbol")
+}
+
+/// Check if a Ruby call's arguments structure represents an unsafe command execution.
+/// Follows strict FAIL-CLOSED principle: unmodeled, dynamic, or shell-executing calls return `true` (UNSAFE).
+/// Only calls with verified non-shell multi-argument or array structures return `false` (SAFE).
+pub fn check_ruby_unsafe_command_injection(
```
juangaitanv: a few on `check_ruby_unsafe_command_injection`: should the shell resolver see through wrapper executables (`system("/usr/bin/env", "sh", "-c", params[:cmd])` classifies as safe today), could it split into callee dispatch / hash filtering / shell-flag scan given lizard flags CCN 27, and how might we keep the ~200 lines of ruby logic out of the language-agnostic file, maybe a `src/language/ruby/` entry?

tests/strictness/language_coverage.rs
```diff
@@ -274,3 +274,114 @@ fn finding_key_set(findings: &[Value]) -> BTreeSet<(String, u64, String, String)
...
+        unsafe_findings.iter().map(|f| (f.line, &f.finding_type, &f.tags)).collect::<Vec<_>>()
+    );
+    println!(
+        "SAFE FINDINGS: {:?}",
+        safe_findings.iter().map(|f| (f.line, &f.finding_type, &f.tags)).collect::<Vec<_>>()
+    );
+
+    // All safe patterns (array literals, multiple arguments, escaped) must be CLEAN!
+    assert!(
+        safe_findings.is_empty(),
```
juangaitanv, line 511: a few on the assertions: could we add a cross-language before/after finding-count check so a ruby change can't silently move JS and Python counts, could the unsafe cases cover the env/options-hash lines 29 to 31 that the `>= 10` aggregate can absorb, and could marker-comment scanning replace the hardcoded line numbers?

juangaitanv (changes requested):
Thanks for pushing this through so many rounds. The fail-closed Ruby rewrite is the right call and reads accurately: unmodeled shapes returning unsafe is the correct default, and the shell-path, namespace, env-hash, and array/splat handling all model argv semantics well. 8 of my 9 earlier threads are cleanly resolved.

One thing I want to own: the fail-closed direction I steered toward across the earlier rounds is part of what pushed logic into the shared extractors, and that's where the two blockers below come from. Flagging it so it doesn't read as a Ruby-only change.

I built `main` and this branch and scanned the repo's own fixtures. Two blockers:

1. **~50 JavaScript findings drop** (CWE-346 postMessage, 14x CWE-922 storage, 2x DOM XSS). The taint-path condition gate now runs on every visited node against every rule's full condition list, so previously-dormant JS conditions activate and non-call nodes fail the `all()`.
2. **7 new Python SQL false positives on `sql_safe_parameterized.py`** (the issue-#35 safe fixture) from the splat / arg-decomposition changes in `common.rs`.

Both likely clear together by scoping the two shared-path changes to the Ruby command-injection condition type rather than the global taint path. Happy to pair on that scoping if useful. Everything in the inline threads below the blockers is optional. Also worth a rerun of `make complexity`: `check_ruby_unsafe_command_injection` trips CCN 27 against the <=15 gate.

Adityakk9031:
@juangaitanv have a look

Adityakk9031:
@juangaitanv and @Ibrahimrahhal please review this

rules/ruby/command_injection.ron
```diff
@@ -25,6 +25,12 @@
                     value: "",
                     condition_type: Some("argument_not_sanitized"),
                     patterns: Some(["Shellwords.escape", "shellescape", "Shellwords.split"])
+                ),
+                (
+                    field: "arguments",
+                    operator: "is_unsafe_command",
+                    value: "",
+                    condition_type: Some("ruby_unsafe_command_injection"),
```
juangaitanv, line 33: the unified taint search path skips `check_ast_conditions`, so argv-safe calls still match the interpolation rule; could we apply the Ruby condition there too?

src/scanner/conditions.rs
```diff
@@ -219,3 +222,340 @@ pub fn check_has_sibling_pattern_condition(
...
+            || lower_arg.starts_with("-enc")
+            || lower_arg.starts_with("/enc")
+            || lower_arg == "-e"
+            || lower_arg == "/e")
+    {
+        return true;
+    }
+
+    false
+}
```
juangaitanv, line 346: clustered POSIX flags such as `-lc` and `-ec` execute command strings but are classified as safe; could we recognize those flags for the applicable shells and add regression coverage?

Adityakk9031:
Thanks @juangaitanv for the detailed review! Both items have been addressed in commit `3905d5f`:

1. **Unified Scan AST Condition Evaluation**: Merged latest `upstream/main` and ensured `check_rule_against_node_with_taint_context` evaluates `check_ast_conditions`. Multi-argument argv calls with interpolation (e.g., `system(ls, #{params[:cmd]})` or `Open3.capture2(echo, #{params[:cmd]})`) are now correctly evaluated by the AST predicate and suppressed as safe.
2. **Clustered POSIX Shell Flags**: Added support in `is_shell_command_flag` / `is_posix_command_flag` for recognizing clustered POSIX flags (e.g. `-lc`, `-ec`, `-elc`, `-ic`) across applicable shells (`sh`, `bash`, `zsh`, `ksh`, `csh`, `tcsh`, `dash`, `ash`, `fish`).
3. **Regression Tests**:
   - Added unit test `ruby_unsafe_command_injection_detects_clustered_posix_flags` in `tests/unit/ast_conditions_tests.rs`.
   - Updated `unsafe.rb` with clustered flag test cases and `safe.rb` with safe non-command shell flags and argv-interpolated forms.
4. **Merge Conflicts & Lints**: Resolved all merge conflicts in `conditions.rs`, `output.rs`, and `django_xss_prevention_tests.rs`. All unit and strictness tests, Clippy (`-D warnings`), and `cargo fmt` pass cleanly.

PR #54, opened 2026-07-22
Add embedded JavaScript DOM XSS scanning

juangaitanv:
## Summary

Add DOM XSS analysis for executable JavaScript embedded in HTML and Django templates.

The scanner now parses eligible inline `<script>` bodies as JavaScript while preserving the original template path and line coordinates. It reuses the canonical frontend DOM XSS taint rules, tracks browser-controlled values such as `location.hash` and promise callback values from `fetch(...).then(...)`, and prefers precise taint findings over the broader HTML fallback finding.

This closes a coverage gap where Sighthound simple analysis missed `document.write` and `innerHTML` flows inside HTML templates. External, empty, non-executable, and unknown script types remain excluded, safe sanitizer cases remain clean, and native JavaScript scanning behavior is unchanged.

Validation covered the two held-out benchmark cases:

- DjangoAt: `location.hash` at line 15 reaches `document.write` at line 20.
- PyGoat: `fetch(*).then` at line 36 reaches `innerHTML` at line 42.

The full Rust suite passed with 211 tests, the acceptance suite passed with 8 scenarios and 31 steps, and the release and HTML-only builds passed. Corpus comparison added only the two expected detection keys; measured overhead was 0.08 seconds for DjangoAt and 0.04 seconds for PyGoat.

## Related issue

None.

## Checklist

- [ ] `make ci` passes locally (the same command CI runs — see CONTRIBUTING.md)
- [ ] Ran `make bootstrap` once so pre-commit/pre-push hooks are active
- [x] Updated docs/rules where relevant


src/scanner/scanning_logic.rs
```diff
@@ -1131,6 +1233,11 @@ impl ScanningLogic {
         let Some(sink_pattern) = ctx.rule_deduplicator.matches_sink_pattern(&node_text) else {
```
yhoztak, line 1283: location.hash is suppressed at the parent node because it has a matching assignment descendant.

The child assignment_expression is not considered actionable for direct source-to-sink detection:
```
fn is_actionable_sink_node(node_kind: &str) -> bool {
    node_kind == "call"
        || node_kind == "expression_statement"
        || node_kind.contains("call_expression")
        || node_kind.contains("invocation")
}
```
Because property assignments also cannot become tainted variables, examples such as these produce no precise result:

```
element.innerHTML = location.hash;
element.outerHTML = document.URL;
element.innerHTML += location.search;

```


The HTML fallback does not help because it only recognizes template-literal patterns. Add regression tests and either make assignment nodes actionable or change the descendant suppression logic.

src/scanner/scanning_logic.rs
```diff
@@ -844,6 +844,105 @@ impl ScanningLogic {
...
+        flow_tracker.record_tainted_variable(
+            parameter,
+            TaintVariableInfo {
+                source_line,
+                source_pattern,
+                source_function: func_name.to_string(),
+                assignment_code,
+            },
+        );
+    }
```
yhoztak, line 950: Promise parameters are recorded by name:
But VariableFlowTracker has one map entry per variable name, not per lexical scope. Reused callback names such as result, response, or data overwrite or inherit each other’s taint. Since source tracking finishes for the whole file before sink processing, this can misattribute both findings or flag a safe callback:

```
fetch("/a").then(result => document.write(result));
Promise.resolve("trusted").then(result => document.write(result));

```

The tracker should use a stable lexical-scope/AST identity, not only the inferred function name and parameter text.

src/scanner/scanning_logic.rs
```diff
@@ -844,6 +844,105 @@ impl ScanningLogic {
...
+            return None;
+        }
+        let source_root = Self::promise_source_root(callee, ctx.source)?;
+        let assignment_code = crate::parser::get_node_text(&source_root, ctx.source);
+        let mut source_chain = assignment_code
+            .chars()
+            .filter(|character| !character.is_whitespace())
+            .collect::<String>();
+        source_chain.push_str(".then");
+        let source_pattern = ctx.rule_deduplicator.matches_source_pattern(&source_chain)?;
```
yhoztak, line 920: Matching is substring-based, and unlike assignment tracking, this path has no sanitizer check. For example, Promise.resolve(DOMPurify.sanitize(location.hash)).then(...) can incorrectly taint the callback parameter. Apply sanitizer evaluation to the resolved expression and test this case.

src/rules.rs
```diff
@@ -152,7 +160,18 @@ impl Rules {
...
+                let javascript_rules = Self::parse_embedded_dir(&RULES_JAVASCRIPT, "JavaScript")?;
+                let embedded_dom_xss_rule = javascript_rules
+                    .iter()
+                    .flat_map(|rules| rules.rules.iter())
+                    .find(|rule| is_frontend_dom_xss_taint_rule(rule))
+                    .ok_or_else(|| {
+                        anyhow::anyhow!("Canonical frontend DOM-XSS taint rule is missing")
+                    })?;
+                all_rules.push(Self { rules: vec![embedded_dom_xss_rule.clone()] });
+            }
```
yhoztak, line 172: `--use-file-rules` loading an HTML-only project never includes that JavaScript taint rule, so users silently receive only the legacy fallback. File and embedded rule modes should have parity, at least when loading the repository’s standard rules.

src/scanner/scanning_logic.rs
juangaitanv: Fixed in 436bbe1. Assignment nodes are now actionable for the direct source-to-sink path, with sources matched against the assigned value only so the sink property cannot self-match. Added regression tests for innerHTML =, outerHTML =, and innerHTML += flows.

src/scanner/scanning_logic.rs
juangaitanv: Fixed in 436bbe1. Embedded JS taint now scopes anonymous callbacks to a positional closure identity (closure@line:column) instead of collapsing into the shared global scope, so reused parameter names cannot inherit taint across callbacks. Regression test covers the fetch vs Promise.resolve("trusted") pair.

src/scanner/scanning_logic.rs
juangaitanv: Fixed in 436bbe1. The resolved promise expression now goes through the same sanitizer evaluation as assignment tracking, so Promise.resolve(DOMPurify.sanitize(location.hash)).then no longer taints the callback parameter. Added a safe-fixture regression test.

src/rules.rs
juangaitanv: Fixed in 436bbe1. It was worse than described: an HTML-only project with --use-file-rules errored with "No taint flow rules found". File-rules mode now loads the canonical DOM-XSS taint rule from rules/javascript for html/django, deduplicated when javascript is also detected, matching embedded-rules behavior. Covered by two new load_rules tests.

src/scanner/scanning_logic.rs
```diff
@@ -1528,8 +1748,10 @@ impl ScanningLogic {
             // when source filtering is enabled.
             "assignment"
             | "call"
+            | "call_expression"
```
leenk7991: can we also pass `is_embedded_javascript` to `should_collect_node` and gate "call_expression" and "argument_assignment_expression" on it

juangaitanv: Done — threaded embedded-JavaScript mode through node collection and limited call_expression and augmented_assignment_expression to embedded scans.

src/scanner/scanning_logic.rs
```diff
@@ -1289,6 +1513,7 @@ impl ScanningLogic {
     ) -> Option<(String, Vec<String>)> {
         let target_var = TaintExpressionUtils::normalize_variable(left_side);
         let mut dependent_vars = CommonUtils::extract_all_variables(right_side);
+        dependent_vars.extend(CommonUtils::extract_simple_variables(right_side));
```
leenk7991: this method `detect_generic_assignment_propagation` is shared by every language, there is no `is_embeddedjs` flag here. `extract_simple_variables` is blunt and will split on `separators` then keep every surviving identifier, so 2 kids of non-vars land in depend_vars.

- `post.name` yields ["post", "name"] then `dependent_vars` is looked up by bare string against `VariableFlowTracker::tainted_variables`, a flat map with no notion of property vs variable.. so an unrelated tainted local named name in the same function taints the target through a flow that doesn't exist.

- `"` isn't a separator, so it only shields the word it's glued to. `q = "SELECT * FROM users WHERE id = " + uid` went from `["uid"]` to `["id", "uid", "users"]`


i checked this  against `tests/test_files/python/`: 172 findings before and after, identical finding sets. nothing regresses today. But the mechanism does fire: `full_path = os.path.join("/uploads", filename)` (`taint_mode_patterns.py:59`) now carries `os`, `path`, and `join` as dependencies, they just dont collide with a tainted name in a corpus this small

juangaitanv: done; separator-based RHS extraction is now limited to embedded JavaScript, with regressions covering member properties and string tokens in non-embedded scans.

src/scanner/vulnerability_scanner.rs
```diff
@@ -486,6 +658,25 @@ impl VulnerabilityScanner {
                 ));
             }
 
+            let has_embedded_dom_xss_rules = !rules.embedded_dom_xss_rules.is_empty();
+            if matches!(detected_language, "html" | "django")
```
leenk7991, line 662: `detect_language_from_path` never returns "django"

src/scanner/modes.rs
```diff
@@ -143,12 +148,25 @@ fn load_rules(cli: &Cli, context: &ScanContext) -> Result<Rules> {
...
 
+/// For HTML/Django, load the canonical frontend DOM-XSS taint rule from the file rules
+/// tree so embedded-script scanning has parity with embedded-rules mode. Returns `None`
+/// for other languages.
+fn load_file_dom_xss_taint_rule(language: &str, base_rules_dir: &str) -> Option<Rules> {
+    if !matches!(language, "html" | "django") {
+        return None;
+    }
+    let javascript_rules_dir = format!("{}/javascript", base_rules_dir);
+    Rules::load_from_directory(&javascript_rules_dir).ok()?.frontend_dom_xss_rule_only()
```
leenk7991, line 167: .ok()? here turns a missing/broken rules/javascript/ into a silent zero-coverage scan that still exits 0

PR #60, opened 2026-07-23
Fix Clippy build failure on Rust 1.91

juangaitanv:
## Summary

- make the AST traversal and import collection helpers private associated functions
- remove unused `&self` receivers that trigger strict Clippy on Rust 1.91
- cover nested AST traversal and nested import collection with regression tests
- preserve behavior without lint suppressions or toolchain changes

## Root cause

Clippy 1.91 reports method receivers used only for recursive calls through `only_used_in_recursion`. The repository runs Clippy with `-D warnings`, so the two warnings fail the build.

This restores strict Clippy compatibility for developers using Rust 1.91 while remaining clean under current Clippy.

Linear: [COR-1692](https://linear.app/corgea/issue/COR-1692/sighthound-build-failure-on-main)

## Validation

- `make check` — 201 tests passed
- `make ci` — all blocking CI gates passed
- `cargo clippy -- -D warnings -D clippy::self_only_used_in_recursion`
- `rustup run 1.91.0 cargo clippy -- -D warnings`
- pre-push Clippy, format, acceptance, and architecture gates

### DIFF
--- a/src/code_type_detector.rs
+++ b/src/code_type_detector.rs
@@ -497,4 +497,20 @@ mod tests {
         let imports = vec!["totally-unrelated-package".to_string()];
         assert_eq!(detector.detect_from_imports(&imports), CodeType::Unknown);
     }
+
+    #[test]
+    fn detect_from_ast_finds_nested_backend_signal() {
+        let source = br#"
+            function loadConfig() {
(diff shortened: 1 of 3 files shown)


yhoztak (approved):
LGTM
confirmed as working now

PR #61, opened 2026-07-24
Remove tracked Python bytecode

juangaitanv:
## Summary

Remove a checked-in CPython 3.13 bytecode cache from the test fixtures. The source script remains tracked, and `.gitignore` already excludes `__pycache__/` and `*.pyc`, so future local runs will not re-add generated bytecode.

## Impact

The repository no longer carries 14,846 bytes of interpreter-specific generated output. Runtime and test behavior do not change.

## Root cause

The bytecode file was committed before the Python cache ignore rules were added. Git continued tracking it after those rules existed.

## Related issue

None.

## Checklist

- [x] Added/updated tests for the change (not applicable: generated artifact deletion)
- [x] Updated docs/rules where relevant (not applicable)

## Verification

- `git ls-files -ci --exclude-standard` returns no paths.
- `make ci` exits 0: strict Clippy, format, audit, 201 tests, acceptance, coverage, advisory complexity/CRAP, and architecture checks.
- Pre-push checks pass.


yhoztak (approved):
LGTM

PR #64, opened 2026-07-28
Fix CSV field escaping

juangaitanv:
## Summary

Fix CSV serialization so every textual field is quoted when it contains a comma, quote, CR, or LF. The previous formatter only escaped the code field, so valid finding values such as a path containing a comma shifted subsequent columns and produced malformed CSV.

Add a focused escaping unit test and a CLI regression test that scans a Python file whose path contains a comma. No dependency was added, and ordinary CSV output remains unchanged.

## Related issue

None.

## Checklist

- [x] Updated docs/rules where relevant (not applicable; no interface or rule change)

### DIFF
--- a/src/scanner/output.rs
+++ b/src/scanner/output.rs
@@ -315,8 +315,8 @@ pub fn print_findings_sarif(findings: &[Finding]) -> Result<()> {
 }
 
 #[cfg(test)]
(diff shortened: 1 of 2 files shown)


yhoztak (approved):
LGTM

PR #66, opened 2026-08-02
fix(scanner): resolve DataFlowTracer stack overflow on cyclic dependencies and rule cache contamination

Adityakk9031:
### Summary

This PR resolves a critical, unlisted engine flaw in Sighthound's static dataflow and taint analysis engine (\DataFlowTracer\):

1. **Stack Overflow Panic on Cyclic Variable Dependencies**:
   - Encountering cyclic variable assignments (e.g. \ = b\, \ = a\, \sink(a)\) previously caused infinite recursion during sink variable resolution, leading to a process crash via stack overflow panic (\STATUS_STACK_OVERFLOW\ / \SIGSEGV\).
   - Fixed by introducing \in_flight_variables\ tracking in \DataFlowTracer::analyze_sink_variable\ to safely break variable dependency resolution cycles early.

2. **Cross-Rule Cache Contamination & Vulnerability Escapes**:
   - \ariable_source_cache\ previously keyed solely on \(file, function, variable)\ without incorporating active rule source patterns. Variable classification from Rule A contaminated Rule B during multi-rule scans, leading to **false negative vulnerability escapes** and false positives.
   - Fixed by keying \ariable_source_cache\ in \DataFlowTracer::analyze_variable_source\ by \(file, function, variable, rule_fingerprint)\, guaranteeing complete cache isolation across rule sets.

### Verification

- Created unit tests in \	ests/unit/dataflow_tests.rs\:
  - \	est_cyclic_variable_dependency_does_not_stack_overflow\: **PASSED**
  - \	est_rule_fingerprint_cache_isolation\: **PASSED**
- All 107 unit, integration, and doc-tests passed with zero errors.


Adityakk9031:

src/scanner/dataflow.rs
```diff
@@ -186,8 +206,13 @@ impl DataFlowTracer {
...
     ) -> VariableSource {
-        let cache_key =
-            (file_path.to_string(), function_name.to_string(), variable_name.to_string());
+        let rule_fingerprint = format!("{:?}", rule_deduplicator);
+        let cache_key = (
+            file_path.to_string(),
+            function_name.to_string(),
+            variable_name.to_string(),
+            rule_fingerprint,
+        );
```
0xjgv: each lookup serializes and retains the full ruleset debug dump; could we compute one compact deduplicator identity and use it in the cache key?

tests/unit/dataflow_tests.rs
```diff
@@ -0,0 +1,117 @@
...
+        cwe_id: None,
+        tags: None,
+    };
+
+    let rules = vec![&rule];
+    let deduplicator = TaintRuleDeduplicator::new(&rules);
+    let mut tracer = DataFlowTracer::new();
+
+    let result = tracer.analyze_sink_variable(path_str, "test_func", "a", "sink", 5, &deduplicator);
+    assert!(matches!(result, AnalysisResult::DefinitelySafe | AnalysisResult::Unknown { .. }));
```
0xjgv: the cycle path returns `Unknown` with a specific reason; should we assert that result so a proven-safe classification cannot pass this regression?

src/scanner/mod.rs
```diff
@@ -1,15 +1,15 @@
...
+pub mod flow_tracker;
 pub mod modes;
 mod multifile_taint;
 mod output;
 mod parser_helper;
 pub mod prefilter;
 mod scan_context;
 mod scanning_logic;
-mod taint_utils;
+pub mod taint_utils;
```
0xjgv: these exports make incomplete scanner internals public; could we move the regression into a `#[cfg(test)]` module and keep the modules private?

Adityakk9031:
Thanks @0xjgv for the great review! All three feedback items have been addressed in commit

src/scanner/dataflow.rs
```diff
@@ -1408,3 +1432,133 @@ impl DataFlowTracer {
...
+
+        let mut tracer = DataFlowTracer::new();
+
+        let result1 =
+            tracer.analyze_sink_variable(path_str, "test_func", "var", "sink1", 3, &dedup1);
+        let result2 =
+            tracer.analyze_sink_variable(path_str, "test_func", "var", "sink2", 3, &dedup2);
+
+        assert!(matches!(result1, AnalysisResult::DefinitelyTainted { .. }));
+        assert!(!matches!(result2, AnalysisResult::DefinitelyTainted { .. }));
```
juangaitanv, line 1757: a leaked in-flight key makes the second call `Unknown`; could we assert its non-cyclic result, then rerun the first rule set and assert `DefinitelyTainted`?

Adityakk9031:
Thanks @juangaitanv for the follow-up review!

1. **Updated Test Assertions**:
   - \
esult2\ is now explicitly asserted as \AnalysisResult::DefinitelySafe\ (verifying non-cyclic evaluation).
   - Added \
esult3\ re-running \dedup1\ after \dedup2\ to verify that the cached hit resolves to \DefinitelyTainted\ without leaking in-flight keys across evaluations.

2. **Merge Conflicts Resolved**:
   - Merged latest \origin/main\ into \ix/dataflow-cycle-and-cache-isolation\ and resolved all merge conflicts in \src/scanner/dataflow.rs\ and \src/scanner/output.rs\.

Adityakk9031:
Updated the test assertion in `test_rule_fingerprint_cache_isolation` in commit `6e0478f`:

- `var = input()` under `dedup2` (where `sources` = `["different_source()"]`) evaluates to `AnalysisResult::Unknown { reason: "unresolved value source: input()" }` because `input()` is unresolvable for `dedup2`.
- `result2` is now explicitly asserted as:
  - `!matches!(result2, AnalysisResult::DefinitelyTainted { .. })` (verifying `dedup1`'s tainted result was not returned from cache)
  - `reason != "cyclic variable dependency"` (verifying `in_flight_variables` did not leak)
- `result3` re-evaluating `dedup1` verifies the cached hit resolves cleanly to `DefinitelyTainted`.

src/scanner/dataflow.rs
```diff
@@ -202,11 +219,13 @@ impl DataFlowTracer {
         sink_line: usize,
         rule_deduplicator: &TaintRuleDeduplicator,
     ) -> VariableSource {
+        let rule_fingerprint = rule_deduplicator.fingerprint();
```
juangaitanv: production shares one deduplicator across rules, so the fingerprint does not isolate rule-specific cache entries. could we scope it to the active rule or derive a sink-specific identity and cover that path?

Adityakk9031:
@juangaitanv have  a look

PR #67, opened 2026-08-15
fix(scanner): enable multi-file cross-taint analysis across all supported languages

Adityakk9031:
## Summary
- Generalizes select_cross_file_targets in src/scanner/multifile_taint.rs to target all active languages in iles_by_language instead of hardcoding Python only.
- Updates uild_import_export_maps to process all supported target languages.
- Generalizes sibling source file lookup and definition pattern matching in src/scanner/dataflow.rs to support non-Python extensions and multiple language function signatures.

### DIFF
--- a/src/scanner/dataflow.rs
+++ b/src/scanner/dataflow.rs
@@ -1102,23 +1102,29 @@ impl DataFlowTracer {
         Some(module_b_path)
     }
 
-    /// Fallback: scan Python files in the same directory as `calling_file` for one that
-    /// defines `function_name`.
-    fn find_in_sibling_python_files(
+    /// Fallback: scan source files with the same extension in the same directory as `calling_file`
+    /// for one that defines `function_name`.
+    fn find_in_sibling_source_files(
         &self,
         calling_dir: &str,
         calling_file: &str,
         function_name: &str,
     ) -> Option<String> {
         let entries = std::fs::read_dir(calling_dir).ok()?;
-        let calling_basename = std::path::Path::new(calling_file).file_name().unwrap_or_default();
+        let calling_path = std::path::Path::new(calling_file);
+        let calling_ext = calling_path.extension().and_then(|e| e.to_str()).unwrap_or("");
+        let calling_basename = calling_path.file_name().unwrap_or_default();
 
         for entry in entries.flatten() {
             let os_file_name = entry.file_name();
             let Some(file_name) = os_file_name.to_str() else {
                 continue;
             };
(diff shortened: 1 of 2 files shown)


Adityakk9031:

src/scanner/multifile_taint.rs
```diff
@@ -327,16 +329,10 @@ impl MultiFileTaintAnalyzer {
...
-                if language == "javascript" || language == "python" {
-                    for file_path in files {
-                        self.analyze_file_for_imports_exports(
-                            file_path,
-                            language,
-                            &rule_deduplicator,
-                        )?;
-                    }
+                for file_path in files {
+                    self.analyze_file_for_imports_exports(file_path, language, &rule_deduplicator)?;
```
juangaitanv: a read or parser failure aborts all cross-file analysis and produces no findings; could we handle failures per file and continue?

src/scanner/dataflow.rs
```diff
@@ -1183,8 +1187,16 @@ impl DataFlowTracer {
...
+            let py_pattern = format!("def {}(", function_name);
+            let js_pattern1 = format!("function {}(", function_name);
+            let js_pattern2 = format!("const {} =", function_name);
+            let js_pattern3 = format!("let {} =", function_name);
+            let fn_pattern = format!("fn {}(", function_name);
+            content.contains(&py_pattern)
+                || content.contains(&js_pattern1)
+                || content.contains(&js_pattern2)
+                || content.contains(&js_pattern3)
+                || content.contains(&fn_pattern)
```
juangaitanv: const and let bindings are treated as functions while body extraction accepts only Python def; could we limit lookup to callable declarations and align body extraction?

src/scanner/multifile_taint.rs
Adityakk9031: @juangaitanv Thanks for the review! Updated `build_import_export_maps` in `src/scanner/multifile_taint.rs` to catch individual file read/parse errors with `if let Err(e) = ... { log::warn!(...); }` and continue processing remaining files instead of failing the entire cross-file analysis.

src/scanner/dataflow.rs
Adityakk9031: @juangaitanv Added `is_function_definition_line` helper that checks only callable declarations (`def`, `async def`, `function`, `async function`, `fn`, `async fn`, `func`, `const x = (`, `let x = (`, etc.) and aligned both `file_contains_function` and `extract_function_body` in `src/scanner/dataflow.rs` to use it consistently.

PR #68, opened 2026-08-15
fix(prefilter): use token-boundary matching for test and migration framework detection

Adityakk9031:
## Summary
- Replaces unbounded substring matching in has_test_patterns and has_migration_patterns with token-boundary aware matching matches_token_or_module.
- Prevents false-positive test file exclusions on legitimate production files that import modules containing substrings of test words (such as \ttestation\, \contest\, \latest\, \astest\, \user_migration\).
- Adds regression unit test cases in \	ests/unit/prefilter_should_scan_tests.rs\.

### DIFF
--- a/src/scanner/prefilter.rs
+++ b/src/scanner/prefilter.rs
@@ -216,7 +216,45 @@ impl PreFilter {
         }
     }
 
-    /// Simple pattern matching for test frameworks
+    /// Helper to check if a token or module pattern matches on identifier boundaries
+    fn matches_token_or_module(text: &str, pattern: &str) -> bool {
+        if pattern.contains('.')
+            || pattern.contains('@')
+            || pattern.contains('/')
+            || pattern.contains('-')
+        {
+            return text.contains(pattern);
+        }
+
+        let mut search_idx = 0;
+        while let Some(pos) = text[search_idx..].find(pattern) {
+            let abs_pos = search_idx + pos;
+            let before_ok = if abs_pos == 0 {
+                true
+            } else {
+                let prev_char = text[..abs_pos].chars().next_back().unwrap();
+                !prev_char.is_alphabetic()
+            };
+
+            let end_pos = abs_pos + pattern.len();
+            let after_ok = if end_pos >= text.len() {
+                true
+            } else {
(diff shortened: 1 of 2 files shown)


src/scanner/prefilter.rs
```diff
@@ -240,25 +278,24 @@ impl PreFilter {
             "org.testng",
         ];
 
-        test_patterns.iter().any(|pattern| imports_text.contains(pattern))
+        test_patterns.iter().any(|pattern| Self::matches_token_or_module(imports_text, pattern))
```
juangaitanv, line 289: vitest, Go testing, and testify no longer match the test patterns, so their imports are scanned. could we retain explicit patterns for these framework roots while preserving the boundary fix?

src/scanner/prefilter.rs
```diff
@@ -240,25 +278,24 @@ impl PreFilter {
...
     }
 
-    /// Simple pattern matching for migration frameworks
+    /// Pattern matching for migration frameworks on token boundaries
     fn has_migration_patterns(&self, imports_text: &str) -> bool {
         let migration_patterns = [
-            "migration",
-            "migrations",
-            "migrate",
```
juangaitanv, line 251: from django.db import migrations no longer matches a migration pattern, so those files are scanned in general mode. could we retain generic migration indicators while treating `_` and digits as identifier characters?

src/scanner/prefilter.rs
Adityakk9031: @juangaitanv Thanks for the review! Added explicit patterns for `"testing"`, `"testify"`, `"vitest"`, and `"github.com/stretchr/testify"` to `test_patterns` while retaining token boundary checking. Added unit test `file_with_testing_or_vitest_import_is_skipped`.

src/scanner/prefilter.rs
Adityakk9031: @juangaitanv Restored `"migration"` and `"migrations"` to `migration_patterns`, and updated `is_identifier_character` to treat `[A-Za-z0-9_]` as non-boundary identifier characters. Now `from django.db import migrations` matches and is skipped, while identifiers like `attestation` or `user_migration` are not falsely matched. Added unit tests for both cases.

PR #69, opened 2026-08-15
fix(codetype): support multiline and side-effect import detection

Adityakk9031:
## Summary
- Adds dot-all flag \(?s)\ and non-greedy matching to the ES \import\ regex in \CodeTypeDetector::extract_imports\ so multiline import blocks are captured accurately.
- Adds regex pattern support for side-effect imports (\import 'package'\).
- Fixes false-positive Frontend classifications for Node/NestJS backend TypeScript files using multiline imports, which previously prevented backend security rules from loading.
- Adds regression unit tests in \src/code_type_detector.rs\.

### DIFF
--- a/src/code_type_detector.rs
+++ b/src/code_type_detector.rs
@@ -257,11 +257,12 @@ impl CodeTypeDetector {
     fn extract_imports(&self, content: &str, _language: &str) -> Vec<String> {
         let mut imports = Vec::new();
 
-        // Extract require() calls
+        // Extract require() calls and ES import statements (including multiline and side-effect imports)
         let require_patterns = vec![
             regex::Regex::new(r#"require\(\s*['"]([^'"]+)['"]\s*\)"#).unwrap(),
-            regex::Regex::new(r#"import\s+.*from\s+['"]([^'"]+)['"]"#).unwrap(),
+            regex::Regex::new(r#"(?s)import\s+.*?from\s+['"]([^'"]+)['"]"#).unwrap(),
             regex::Regex::new(r#"import\s*\(\s*['"]([^'"]+)['"]\s*\)"#).unwrap(),
+            regex::Regex::new(r#"import\s+['"]([^'"]+)['"]"#).unwrap(),
         ];
 
         for pattern in require_patterns {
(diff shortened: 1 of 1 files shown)


src/code_type_detector.rs
```diff
@@ -257,11 +257,12 @@ impl CodeTypeDetector {
     fn extract_imports(&self, content: &str, _language: &str) -> Vec<String> {
         let mut imports = Vec::new();
 
-        // Extract require() calls
+        // Extract require() calls and ES import statements (including multiline and side-effect imports)
         let require_patterns = vec![
             regex::Regex::new(r#"require\(\s*['"]([^'"]+)['"]\s*\)"#).unwrap(),
-            regex::Regex::new(r#"import\s+.*from\s+['"]([^'"]+)['"]"#).unwrap(),
+            regex::Regex::new(r#"(?s)import\s+.*?from\s+['"]([^'"]+)['"]"#).unwrap(),
```
juangaitanv: the DOTALL match can capture a later `from` inside unrelated source text; could we constrain it to one import declaration and add a negative regression test?

src/code_type_detector.rs
```diff
@@ -257,11 +257,12 @@ impl CodeTypeDetector {
...
             regex::Regex::new(r#"import\s*\(\s*['"]([^'"]+)['"]\s*\)"#).unwrap(),
+            regex::Regex::new(r#"import\s+['"]([^'"]+)['"]"#).unwrap(),
```
juangaitanv, line 272: a Go import such as `github.com/reactivex/rxgo` can match the `react` signature and override the Backend fallback; should we limit this extraction to applicable languages?

src/code_type_detector.rs
Adityakk9031: @juangaitanv Thanks for catching this! Constrained the import regex using `[^;]+?` so it cannot capture across statement boundaries / semicolons into unrelated `from` occurrences. Added negative regression test `import_regex_does_not_capture_across_statement_boundaries`.

src/code_type_detector.rs
Adityakk9031: @juangaitanv Thanks! Restricted `extract_imports` in `src/code_type_detector.rs` to JavaScript / TypeScript languages (`javascript`, `typescript`, `tsx`, `jsx`, `js`, `ts`), ensuring non-JS languages like Go or Python do not extract false JS framework signatures and cleanly retain their backend classification. Added regression test `non_js_imports_do_not_override_backend_default`.

PR #70, opened 2026-08-15
fix(rules): do not hardcode js extensions in apply_centralized_exclusions

Adityakk9031:
\, \.jsx\, \.ts\, \.tsx\])\.
- This previously caused rules loaded for Python, Java, Go, C#, Ruby, PHP, and generic search rules to be rejected during file scanning when evaluated by \
ule_applies_to_file\.
- Sets \extensions: None\ so file extension matching remains unrestricted for generic rules.
- Adds regression unit test in \	ests/unit/exclusion_patterns_tests.rs\.

### DIFF
--- a/tests/unit/exclusion_patterns_tests.rs
+++ b/tests/unit/exclusion_patterns_tests.rs
@@ -47,4 +47,39 @@ mod exclusion_patterns_tests {
         assert_eq!(p.get_patterns("backend"), Vec::<String>::new());
         assert_eq!(p.get_patterns("common"), Vec::<String>::new());
     }
+
+    #[test]
+    fn apply_centralized_exclusions_does_not_hardcode_js_extensions() {
+        let mut rules = sighthound::rules::Rules {
+            rules: vec![sighthound::UnifiedRule {
+                id: Some("test-rule".to_string()),
+                name: Some("Generic Rule".to_string()),
+                description: None,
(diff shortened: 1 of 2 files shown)


Adityakk9031:

tests/unit/exclusion_patterns_tests.rs
```diff
@@ -47,4 +47,39 @@ mod exclusion_patterns_tests {
...
+                conditions: None,
+                tags: None,
+                cwe_id: None,
+                message: None,
+            }],
+        };
+
+        rules.apply_centralized_exclusions(&patterns(), "backend");
+        let file_types = rules.rules[0].file_types.as_ref().unwrap();
+        assert!(file_types.extensions.is_none());
```
juangaitanv: the synthesized fields do not exercise the regressed applicability behavior; could we test `rule_applies_to_file` for a non-JS path and an excluded path?

Adityakk9031: @juangaitanv Thanks for the review! Updated the unit tests in `tests/unit/exclusion_patterns_tests.rs` to explicitly exercise `rule_applies_to_file` on non-JS paths (`src/main.py`, `src/service.go` -> true) and excluded paths (`src/app.min.js`, `src/handler_test.go` -> false). All tests passing.

PR #71, opened 2026-08-15
fix(ast_provenance): resolve duplicate function definitions by usage line span

Adityakk9031:
## Summary
- Stores all definitions for a given function name along with their AST line spans in \PythonAstProvenance\.
- Updates \
esolve_variable_reaching\ to match the sink's \usage_line\ against each definition's line span when multiple methods in a file share the same bare name (e.g. \get\, \post\, \
un\, \handle\ across multiple classes).
- Resolves false-negative vulnerability drops where class methods previously collided in \mbiguous\ and were ignored by \DataFlowTracer\.
- Adds regression unit tests in \src/scanner/ast_provenance.rs\.

### DIFF
--- a/src/scanner/ast_provenance.rs
+++ b/src/scanner/ast_provenance.rs
@@ -70,17 +70,13 @@ pub(crate) enum AstResolution {
 struct FunctionFacts {
     parameters: Vec<String>,
     assignments: Vec<AssignmentFact>,
+    span: (usize, usize),
 }
 
 #[derive(Debug)]
 struct FileFacts {
-    /// First definition in document order wins, matching the text path.
-    functions: std::collections::BTreeMap<String, FunctionFacts>,
-    /// Bare names defined more than once in the file (e.g. `run` on two
-    /// classes). The caller resolves a variable by the sink's *bare* function
-    /// name, so for such names we cannot know which definition is meant and
-    /// refuse to guess — see [`PythonAstProvenance::resolve_variable`].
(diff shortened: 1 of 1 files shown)


Adityakk9031:

src/scanner/ast_provenance.rs
```diff
@@ -122,15 +118,27 @@ impl PythonAstProvenance {
...
+                    if matching.len() == 1 {
+                        matching[0]
+                    } else {
+                        return Some(AstResolution::Ambiguous);
+                    }
+                } else {
+                    return Some(AstResolution::Ambiguous);
+                }
+            }
+        };
```
juangaitanv, line 141: when the selected definition has no local facts, the text fallback can resolve a different same-named definition; could we return `AstResolution::Ambiguous` instead?

Adityakk9031: @juangaitanv Thanks for the feedback! Updated `resolve_variable_reaching` in `src/scanner/ast_provenance.rs` so that when `function_list.len() > 1` (duplicate function definitions exist in the file) and the selected definition lacks local facts (no matching assignments or parameters for the variable), it returns `Some(AstResolution::Ambiguous)`. This prevents the text fallback from incorrectly attributing another same-named definition's facts. Added regression test `duplicate_function_names_without_local_facts_resolve_as_ambiguous`.

Adityakk9031:
@juangaitanv and @Ibrahimrahhal  have look on all the open prs

PR #73, opened 2026-08-19
CWE-95 + frontend/template XSS precision

leenk7991:
## Summary
Stop Sighthound from reporting DOM APIs as CWE-95 eval injection, and cut Django/HTMX XSS noise without dropping real DOM XSS.

**Problem:** CWE-95 findings were innerHTML/htmx/DOMPurify writes, rated Critical. HTML-only files failed combined mode (`No taint flow rules found`) so Django `|safe` never ran. `vm.runInNewContext(userInput)` was silent (taint-only sink; frontend sources are DOM, not generic params; search prefilter dropped regex-escaped `vm.runIn*` names). `escapeHtml` (`textContent` then read `innerHTML`) was CWE-79.

**Root cause:** CWE-95 sink list too broad (any HTML write ≈ eval). Combined mode required taint rules even when the language only has search rules. HTML language did not visit `text` nodes. No search-mode `vm` pattern that survived the call-name prefilter.

## Changes
- eval sinks: eval / Function / setTimeout(string literal) / setInterval(string literal) / vm.runIn* only — not HTML writes, not `setTimeout(handler, n)`
- Search timers match a quote after `(` (`setTimeout('…')`); identifier callbacks stay quiet. Taint still flags a timer when the first arg is a known source
- Dummy `vm.runInNewContext=` (and This/Context) names so the search prefilter keeps `vm.runIn*` (escaped regex dots are not a substring of the call name)
- XSS: taint + unsafe HTML sink; autoescape / DOMPurify / json_script / hx-swap allow
- `escapeHtml`: textContent write then innerHTML read is a sanitizer, not CWE-79
- Django: `|safe` / `|mark_safe` on request.GET/POST/COOKIES only
- Combined mode skips taint when there are 0 taint rules (HTML/Django search still runs)
- HTML language visits `text` nodes so `|safe` matches when `.html` auto-detects as `html`
- Backend SSTI (`ejs.render` / handlebars / mustache) is no longer CWE-95 (eval-family only; CWE-94 follow-up)
- `build.rs` reruns compile when `rules/` change
- Ignore `/sighthound_release/` (local platform export; do not commit the binary)
- fixtures: TN innerHTML helper, escapeHtml, setTimeout/setInterval callbacks and `setTimeout(handler)`; TP eval / Function / string-literal timers / vm / `|safe`

## Test plan
- [x] existing JS XSS TPs still fire
- [x] new TNs do not fire CWE-95 (`escapeHtml`, DOMPurify, `template.innerHTML`, `htmx.trigger`, callback timers)
- [x] `escapeHtml` is not CWE-79
- [x] `innerHTML = location.hash` remains CWE-79
- [x] string `setTimeout(userInput)` is CWE-95; `setTimeout(function () {…})` / `setTimeout(handler)` is not
- [x] string `setInterval(userInput)` is CWE-95; `setInterval(function () {…})` is not
- [x] `eval` / `new Function` / `Function('return '+x)` / `eval(location.hash)` / `vm.runInNewContext(userInput)` are CWE-95
- [x] Django `|safe` and `| mark_safe` on request data are CWE-79; autoescape / `json_script` / `hx-swap` are not
- [x] scanning the same template as language `html` still flags `|safe`
- [x] `cargo test --test strictness_tests -- cwe95_xss_sink_precision`

E2E (fixture Django + JS app, this branch vs `main`):

- [x] `eval` / `Function` reported as CWE-95 (were CWE-94 on `main`)
- [x] `setTimeout(userInput)` with a string argument reported as CWE-95 (missing on `main`)
- [x] `innerHTML = location.hash` still reported as CWE-79
- [x] `escapeHtml` / `createElement` / parse-only `innerHTML` helpers no longer reported as CWE-95
- [x] `setTimeout(function () {…})` callback not CWE-95
- [x] Django `|safe` / `|mark_safe` on request data reported as CWE-79 (missing on `main`)

## Checklist

- [x] Updated docs/rules where relevant

linear ticket: https://linear.app/corgea/issue/COR-1802/


src/scanner/utils.rs
```diff
@@ -609,6 +609,71 @@ impl AstUtils {
...
+        for (i, c) in args.char_indices() {
+            match c {
+                '(' => depth += 1,
+                ')' => {
+                    depth -= 1;
+                    if depth == 0 {
+                        let inner = args.get(1..i).unwrap_or("").trim();
+                        let after = args.get(i + 1..).unwrap_or("").trim_start();
+                        return after.starts_with("=>")
+                            || Self::starts_with_timer_callback_arg(inner);
```
juangaitanv, line 678: parenthesized timer IIFEs are classified as callbacks and skip CWE-95 findings; could we treat a following `(` as an invocation?

rules/javascript/frontend_security.ron
```diff
@@ -144,6 +147,49 @@
...
+            mode: "search",
+            // Dummy `=` names pass the call-name prefilter (escaped regex dots
+            // are not a substring of `vm.runInNewContext`).
+            patterns: Some([
+                "vm.runInNewContext=",
+                "vm.runInThisContext=",
+                "vm.runInContext=",
+                "regex:vm\\.runIn(?:New|This)?Context\\(\\s*[A-Za-z_$]",
+                "regex:vm\\.runIn(?:New|This)?Context\\(\\s*['\"]"
+            ]),
```
juangaitanv, line 190: constant timer strings and trusted compiled scripts match as Critical CWE-95. could we require dynamic or untrusted input and leave identifiers to taint analysis?

rules/backend_javascript/backend_security.ron
```diff
@@ -474,25 +474,14 @@
             ]),
             
             sinks: Some([
-                // Direct code execution
+                // Eval injection only. Template engines / dynamic require are not CWE-95.
```
juangaitanv, line 477: removing these sinks leaves request-sourced server template injection uncovered. should we retain them under CWE-94 or include a replacement rule?

src/language.rs
```diff
@@ -548,6 +569,7 @@ impl LanguageSupport for HTMLLanguage {
         // names like `th:utext`, `th:replace`, or tag names like `textarea`
         // resolve as the matchable "function" name for search rules.
         match node.kind() {
+            "text" => django_template_name_from_text(get_node_text_slice(node, source)),
```
juangaitanv, line 595: Django filters in attribute values do not reach the safe-filter rule. could we extract template tokens from attribute values as well?

rules/html/django.ron
```diff
@@ -0,0 +1,37 @@
...
+            category: Some("xss"),
+            mode: "search",
+            patterns: Some([
+                "django-safe-filter-sink=",
+                "*request.GET*|safe*",
+                "*request.POST*|safe*",
+                "*request.COOKIES*|safe*",
+                "*request.GET*|mark_safe*",
+                "*request.POST*|mark_safe*",
+                "*request.COOKIES*|mark_safe*",
```
juangaitanv: nit: full-node matching combines adjacent Django expressions; could we bound wildcard matches to one `{{ ... }}` expression and add a negative regression?

rules/javascript/frontend_security.ron
```diff
@@ -144,6 +147,51 @@
...
+            // Gate `=` forces full-call matching. First arg must be a concatenated
+            // or interpolated string, not a constant (`setTimeout('tick')`) or
+            // identifier callback (`setTimeout(handler, n)`). Identifiers stay with taint.
+            patterns: Some([
+                "settimeout-eval-sink=",
+                "regex:setTimeout\\(\\s*['\"][^'\"]*['\"]\\s*\\+",
+                "regex:setTimeout\\(\\s*`[^`]*\\$\\{",
+                "regex:setInterval\\(\\s*['\"][^'\"]*['\"]\\s*\\+",
+                "regex:setInterval\\(\\s*`[^`]*\\$\\{"
+            ]),
```
juangaitanv, line 168: nit: qualified `window.setTimeout` and `setInterval` calls are filtered out before matching; could we accept qualified timer callees in the prefilter and add regression coverage?

src/scanner/scanning_logic.rs
```diff
@@ -2220,6 +2239,12 @@ impl ScanningLogic {
         if !Self::rule_pattern_matches_node(rule, &node_text) {
             return None;
         }
+        let finding_type = rule.get_finding_type().to_lowercase();
+        if (finding_type.contains("xss") || rule.cwe_id.as_deref() == Some("cwe-79"))
+            && crate::scanner::utils::AstUtils::is_textcontent_escape_innerhtml_read(&node_text)
+        {
+            return None;
+        }
```
juangaitanv, line 2250: the escape-helper suppression covers unrelated sinks inside script elements; could we scope it to the relevant expression or language and add a regression combining textContent with outerHTML?

src/common.rs
```diff
@@ -66,22 +66,59 @@ impl CommonUtils {
...
+    fn suffix_looks_like_dom_assignment(after: &str) -> bool {
+        let after = after.trim_start();
+        if after.starts_with("+=") || (after.starts_with('=') && !after.starts_with("==")) {
+            return true;
+        }
+        let Some(rest) = after.strip_prefix("as ") else {
+            return false;
+        };
+        rest.find(')')
+            .is_some_and(|paren| Self::suffix_looks_like_dom_assignment(&rest[paren + 1..]))
```
juangaitanv, line 124: logical assignments bypass DOM sink matching; could we recognize ||=, &&=, and ??= and add a tainted-intermediate regression?

src/scanner/utils.rs
```diff
@@ -612,6 +612,76 @@ impl AstUtils {
...
+        if !sink_pattern.contains("setTimeout") && !sink_pattern.contains("setInterval") {
+            return false;
+        }
+        let lower = node_text.to_ascii_lowercase();
+        ["settimeout", "setinterval"].iter().any(|name| {
+            lower.find(name).is_some_and(|idx| {
+                let after_name = lower[idx + name.len()..].trim_start();
+                after_name
+                    .strip_prefix('(')
+                    .is_some_and(|args| Self::starts_with_timer_callback_arg(args.trim_start()))
```
juangaitanv: a callback in one timer suppresses the whole expression, hiding a later sink; could we scope suppression to the matched call or inspect every timer occurrence?

rules/html/django.ron
```diff
@@ -0,0 +1,30 @@
...
+        (
+            id: Some("html-django-safe-request-xss"),
+            name: Some("Django |safe on request data"),
+            category: Some("xss"),
+            mode: "search",
+            patterns: Some([
+                "django-safe-filter-sink=",
+                "|safe=",
+                "|mark_safe=",
+                "{{=",
```
juangaitanv: literal {{= matching flags ordinary template syntax without request data or an unsafe filter; could we gate it on both conditions or remove the literal alternative?

juangaitanv (approved):
lgtm

PR #75, opened 2026-08-25
Add HTML security rules with markup search semantics

juangaitanv:
**Stacked on #74.** This branch sits on `fix-html-taint-regression` (`af1443e`) and must merge **after** it. A diff against `main` will also show #74's taint fix; the 12 files below are this PR's own.

### What this adds

12 HTML security rule IDs in a new `rules/html/html_security.ron`, plus a minimal AST-vs-Markup boundary so HTML and Django match search-mode rules against node text rather than resolved function names.

- `src/language.rs` — new `SearchSemantics { Ast, Markup }`; `LanguageSupport::search_semantics()` defaults to `Ast`; HTML and Django override to `Markup`.
- `src/scanner/scanning_logic.rs` — markup prefiltering and final matching against node text, line attribution to the first positive match, and tightest-span dedup for nested markup findings.
- `src/scanner/conditions.rs` — `node_kind` condition arm.
- `src/rules.rs` — `first_positive_match_range` for line attribution.
- `tests/strictness/html_security.rs` (new), registered in `tests/strictness/main.rs`.
- `tests/strictness/helpers.rs` / `language_coverage.rs` — `run_cli_raw`, `run_cli_json`, `sighthound_binary` promoted from private-in-`language_coverage` to `pub` in `helpers` so the new suite can drive the CLI. Pure relocation; no coverage lost.

### Why it depends on #74

Before #74, `src/scanner/modes.rs` hard-errored when a language had zero taint-mode rules. HTML has none, so every HTML scan exited 1 with zero findings.

Verified rather than assumed: revert `modes.rs` to its pre-#74 state and `html_security_rules_fire_in_default_mode` fails with `Error: No taint flow rules found`. With #74 in the base it passes. This PR adds **no** hunk to `modes.rs`.

### Review notes

- **Shared surface with #74**: `tests/strictness/helpers.rs` and `tests/strictness/language_coverage.rs`. If #74 is amended or force-pushed, the helper move here needs re-deriving.
- **`include_dir!` staleness**: rules are embedded at compile time via `include_dir!` in `src/rules.rs`, and there is no `build.rs`. Editing a `.ron` with a warm `target/` yields zero findings until you `touch src/rules.rs`. Flagging it so the next person loses less time than I did.
- **Pre-existing gap**: `call_node_types()` omits `self_closing_tag`, so rules do not fire on self-closing elements. Not introduced here, not fixed here.
- **Dead code**: `"element"` in `call_node_types()` never produces findings — `get_function_name` resolves `tag_name` via `CommonUtils::find_child`, which walks direct children only, and per tree-sitter-html's node-types an `element` has no direct `tag_name`.

### Not covered by tests

- Django running `Markup`. The override exists; no test exercises it.
- The `.expect()` in the markup line-attribution block. Unreachable by construction (`markup == true` implies `node_text.is_some()`), but unasserted.
- The `get_finding_type()` → `"vulnerability"` fallback.

### Relationship to #9

#9 carries the same 12 rule IDs from the same lineage; this PR supersedes that portion. #9's sql/xml/properties/config rules, its parser-free text-scanning engine, and its `unless` exclusion mechanism are **not** included here — all were on this chunk's exclusion list. #9 stays open for that work.

### Verification

- `cargo test --test strictness_tests` — 47 passed
- `make test-unit` — 261 passed
- `cargo fmt --all -- --check` — clean
- Fusion contract (`sighthound --output-format json <path>`): positive fixture → exit 0, 12 findings at lines 7, 11, 13, 15, 17, 19, 21, 23, 25, 27, 29, 31, empty stderr; safe fixture → `[]`
- `--taint-analysis --output-format json` → stdout is `[]`, notice routed to stderr via `ui::warn`

### DIFF
--- a/rules/html/html_security.ron
+++ b/rules/html/html_security.ron
@@ -1,0 +1,248 @@
+(
+    rules: [
+        // Generic HTML security patterns: dangerous markup that is unsafe on its own,
+        // independent of any template engine. Twelve rules, all `mode: "search"`, all
+        // scoped to `.html` / `.htm`.
+        //
+        // ENGINE NOTE: `.html` files are matched with `SearchSemantics::Markup`
+        // (src/language.rs), so both the prefilter and the final match run against the
+        // full text of each candidate node — `attribute`, `start_tag`, `script_element`
+        // and `element` — rather than against the resolved "function" name. That is what
+        // lets a pattern span two attributes (`html-xss-011`) or sit inside a `<script>`
+        // body (`html-xss-012`, `html-xss-013`, `html-sec-001`), neither of which the
+        // name-based prefilter can reach.
+        //
+        // PATTERN FORM: every pattern must be a plain literal (no `\`, no `*`) or
+        // `regex:`-prefixed. `rules::first_positive_match_range` resolves a byte range
+        // only for those two forms, and that range is what lands a finding on the source
+        // line the offending markup actually sits on. A glob or escaped pattern would
+        // still match but would silently fall back to heuristic line attribution.
+        //
+        // SCOPING: `html-xss-012` and `html-xss-013` carry a `node_kind` condition
+        // pinning them to `script_element`. Without it, `-013` would double-report the
+        // `onclick="eval(...)"` lines already covered by `-004`.
+        (
+            id: Some("html-xss-001"),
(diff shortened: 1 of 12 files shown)


yhoztak (approved):
LGTM

PR #76, opened 2026-08-26
Require value-interpolation markers in html_security.ron sinks

juangaitanv:
## Summary

`rules/html/html_security.ron` (12 rules, merged in #75) matched the *shape* of a
`javascript:`/inline-event payload rather than any evidence of attacker-controlled
data. This PR requires a value-interpolation marker (`{{`, `${`, or `<%=`) inside the
sink for the four rules that had none, drops one rule that became redundant, and fixes
one rule's unbounded regex.

## Corpus results

**Before/after numbers are not directly comparable** — see divergence note below.

- **My baseline** (unmodified `main` @ `8320bf8`, 1,817 deduped real-world `.html`/`.htm`
  files): **5 findings from `html_security.ron`, across 5 files, 0 true positives.**
  - `html-xss-001` × 2: Django's own `admindocs/templates/admin_doc/bookmarklets.html`
    (`{% url %}` inside a `javascript:` href — a control tag, not interpolation), and
    doghouse's `_upgrade_required.html` (`href="javascript:history.back()"`).
  - `html-xss-014` × 2: two bundled/minified JS files (Playwright's htmlReport,
    CLIProxyAPI's management page) — the unbounded `.*\$\{` matched from deep inside a
    `<script>` block to an unrelated `${` far downstream.
  - `html-xss-012` × 1: **a genuine true positive**, see below. (Rule 012 is untouched
    by this PR — it already required a taint keyword.)
- **After this PR** (same 1,817-file corpus): **1 finding from `html_security.ron`,
  1 file, and it is the same true positive.** The four false positives above are gone.
  Confirmed by re-running the fixed binary over the identical corpus.

**Divergence from the task's stated baseline** (49 findings / 16 files / 0 TPs): my
reconstruction does not reproduce that number. Scanning the same corpus with an
unmodified `main` binary gives 5 findings / 5 files from `html_security.ron`, not 49/16.
This machine's caches (`~/.cache/uv`, `~/Library/Caches`) and checkouts churn over time,
so file population differs from whenever the original 1,665-file study ran; I did not
have access to that original file list to diff against. I'm reporting my own
independently-measured before/after pair rather than presenting the prior 49/16 figure
as something I reproduced.

### True positive found (rule html-xss-012, untouched by this PR)

`vuln-benchmarks/.vulnbench/e2e-corpus/realvuln-djangoat/template/users/signup.html:20`:

```js
var hashParam = location.hash.split("#")[1];
...
var paramValue = decodeURIComponent(hashParam.split('=')[1]);
if (paramValue != "undefined") {
    document.write("<OPTION value=3>" + paramValue + "</OPTION>");
}
```

`location.hash` is attacker-controlled (URL fragment), decoded and written directly into
the DOM via `document.write` with no escaping — a genuine DOM-based XSS (CWE-79). An
attacker-crafted fragment (e.g. breaking out of the `value="..."` context) executes in
the victim's browser. This file lives under `corgea/vuln-benchmarks`, a benchmark corpus
of deliberately-vulnerable apps not on the task's named exclusion list (only
pygoat/vulpy/lets-be-bad-guys/fusion-benchmarks were named) — kept per the literal
exclusion list, and it happens to be where the one real finding came from.

### Five must-fire / must-not-fire checks — all pass

| Case | Result |
|---|---|
| `<a href="javascript:{{ user_var }}">` | fires |
| `<a href='javascript:${xss}'>` | fires |
| `<a href="javascript:void(0)">` | does not fire |
| `<a href="javascript:history.back()">` | does not fire |
| Django admindocs bookmarklet (`{% url %}` inside `javascript:`) | does not fire |

## Design change

Payload-shape matching → value-interpolation-marker-in-sink:

- **html-xss-001/002/019** (`javascript:` in `href`/`action`/`src`): now require
  `{{`, `${`, or `<%=` after `javascript:`, bounded to the attribute's own quote
  character (kept the `regex:` prefix so `first_positive_match_range` still resolves an
  exact byte range instead of falling back to heuristic line attribution — see the
  file's own header, originally at lines 15-19, not 19-23 as the task description said).
  001/002 keep a third, bounded unquoted-attribute pattern (the original had one); 019
  gets only the two quoted variants, matching what the original actually had (it never
  had an unquoted variant).
- **html-xss-014** (template syntax in `onclick`/`onmouseover`/`onerror`/`onload`/
  `onchange`): the prior pattern (`onclick\s*=.*\$\{`, no bound) had no `[^"']`-style
  cap, unlike every sibling rule. Two real-corpus firings matched 108,788 and 36,212
  characters. Fixed by bounding to the attribute's own quote character, per quote
  variant (`onclick\s*=\s*"[^"]*\$\{` / `onclick\s*=\s*'[^']*\$\{`, ×5 handlers = 10
  patterns — the `regex` crate has no backreferences, so a single `["']`-class pattern
  can't express "same quote that opened").
- **html-xss-004** (any onclick/onmouseover/.../onsubmit handler containing `eval(`):
  removed. It carried no taint signal at all — not even a template marker — so once
  `-014` requires an actual interpolation marker, keeping `-004` around is strictly less
  precise for the same attributes. (Its match set is not literally a subset of -014's:
  `onclick="eval(userInput)"` has no `${`. It's dropped because it has zero signal, not
  because -014 now subsumes it.)

`{%` is deliberately excluded from the marker set: it's a Django/Jinja *control* tag
(loops, conditionals, `{% url %}`), not value interpolation. Django's own admindocs
bookmarklet template embeds `{% url %}` inside a `javascript:` href and is not
exploitable — confirmed both in the corpus (see above) and as a permanent fixture case
in `html_security_safe.html`.

## Out of scope (deliberately)

- **014's `{{ }}` expansion**: not made. Measured separately against doghouse's 84
  attribute-scoped `{{ }}` occurrences — only ~4 are exploitable; 29 of 33 distinct
  expressions are PKs, UUIDs, loop counters, or already `|escapejs`-filtered.
  Interpolating a row's PK into an `onclick` is idiomatic Django; matching on `{{ }}`
  there would be a net precision loss.
- **012/013's keyword-heuristic redesign** (`document.write(...)`/`eval(...)` gated on
  `user|input|data|param|location|${` rather than a real taint marker): untouched. 012's
  one corpus firing in this PR happens to be a true positive, but the rule's design
  (a keyword allowlist, not actual taint tracking) is unchanged and could still be
  imprecise elsewhere.
- Rules 008, 010, 011, 020, html-sec-001: untouched, as instructed.
- A same-pathology cluster in the **separate, untouched `xss.ron`** pack (12 findings,
  "Assigning a dynamic template literal to innerHTML/outerHTML...", whole-`<script>`-body
  spans on the same corpus) looks like the same unbounded-span issue -014 had, in a
  different rule not covered by this PR. Flagging as follow-up material, not fixed here.

## Test changes

- `tests/test_files/html/html_security_positive.html`: the 001/002/019 cases now carry
  `{{ user_var }}` / `${xss}` markers instead of bare `javascript:alert(1)`; the 004 case
  is removed; line numbers for everything after it shift up by 2 (18→11 findings total).
- `tests/test_files/html/html_security_safe.html`: gained three permanent regression
  cases — `javascript:void(0)`, `javascript:history.back()`, and the full-fidelity
  Django admindocs bookmarklet payload.
- `tests/strictness/html_security.rs`: `EXPECTED` table updated for the new lines/
  snippets (verified against actual `--output-format json` output, not hand-predicted);
  doc comments' rule/attribute counts corrected (also fixed a pre-existing arithmetic
  error in the "attribute-level vs tag-level" count: was "nine" for 8 attribute + 4
  tag-level = 12, now correctly "seven" for 7 + 4 = 11).
- `tests/unit/pattern_matching_tests.rs`: `html-xss-004` removed from the expected rule-
  id set, count 12 → 11, test renamed to `..._eleven_rule_ids`.

## Test suite results

`cargo test --test strictness_tests` (51 passed), `cargo test --test unit_tests` (109
passed), `make test-unit` (268 passed total across all harnesses), `cargo fmt --all --
--check` (clean). All green.

https://claude.ai/code/session_011JyE6ov9cL7KDWUFXiaJ5X

### DIFF
--- a/rules/html/html_security.ron
+++ b/rules/html/html_security.ron
@@ -1,7 +1,7 @@
 (
     rules: [
         // Generic HTML security patterns: dangerous markup that is unsafe on its own,
-        // independent of any template engine. Twelve rules, all `mode: "search"`, all
+        // independent of any template engine. Eleven rules, all `mode: "search"`, all
         // scoped to `.html` / `.htm`.
         //
         // ENGINE NOTE: `.html` files are matched with `SearchSemantics::Markup`
@@ -19,17 +19,34 @@
         // still match but would silently fall back to heuristic line attribution.
         //
         // SCOPING: `html-xss-012` and `html-xss-013` carry a `node_kind` condition
-        // pinning them to `script_element`. Without it, `-013` would double-report the
-        // `onclick="eval(...)"` lines already covered by `-004`.
+        // pinning them to `script_element`. Without it, both patterns would also match
+        // against `attribute`/`start_tag` node text exposed by markup search semantics
+        // (e.g. an `onclick` handler whose value happens to contain `document.write(` or
+        // `eval(` text), firing outside any actual `<script>` body.
+        //
+        // PRECISION (2026-08): `html-xss-001`/`-002`/`-019` (javascript: URL sinks) and
+        // `-014` (template syntax in inline event handlers) were measured against 1,817
+        // deduped real-world HTML files and produced only false positives — the shape of
+        // a `javascript:` payload or an unbounded `.*` scan is not evidence of
+        // attacker-controlled data. All four now require a value-interpolation marker
+        // (`{{`, `${`, or `<%=`) inside the sink, bounded to the attribute's own quote
+        // character. `{%` is deliberately excluded: it is a Django/Jinja *control* tag
+        // (loops, conditionals, `{% url %}`), not value interpolation — Django's own
+        // admindocs bookmarklet template embeds `{% url %}` inside a `javascript:` href
+        // and is not exploitable. `html-xss-004` (any onclick/onerror/... handler
+        // containing `eval(`) was removed outright: it carries no taint signal at all —
+        // not even a template marker — so it is strictly less precise than the bounded
+        // `-014`, which now covers the same attributes with an actual interpolation
+        // check.
         (
             id: Some("html-xss-001"),
             name: Some("JavaScript URL in href Attribute"),
             category: Some("xss"),
             mode: "search",
             patterns: Some([
-                "href=\"javascript:",
-                "href='javascript:",
-                "href=javascript:",
+                "regex:(?i)href\\s*=\\s*\"javascript:[^\"]*(?:\\{\\{|\\$\\{|<%=)",
+                "regex:(?i)href\\s*=\\s*'javascript:[^']*(?:\\{\\{|\\$\\{|<%=)",
+                "regex:(?i)href\\s*=\\s*javascript:[^\\s>]*(?:\\{\\{|\\$\\{|<%=)",
             ]),
             finding_type: Some("Cross-Site Scripting (XSS)"),
             severity: Some("High"),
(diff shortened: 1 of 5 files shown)


yhoztak (approved):
LGTM

